PR #63, opened 2015-07-16
[ENH] Switch regressions to use the pandas implementation

twiecki:
There were several different methods to calculate regressions
being used. We should use a consistent method everywhere
and Pandas does fast rolling regressions with all sorts of
bells and whistles.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -90,20 +90,20 @@ def plot_rolling_risk_factors(
     rolling_risk_multifactor = timeseries.rolling_multifactor_beta(
         returns,
         risk_factors.loc[:, ['SMB', 'HML', 'UMD']],
-        rolling_window=rolling_beta_window)
+        window=rolling_beta_window)
 
     rolling_beta_SMB = timeseries.rolling_beta(
         returns,
         risk_factors['SMB'],
-        rolling_window=rolling_beta_window)
+        window=rolling_beta_window)
     rolling_beta_HML = timeseries.rolling_beta(
         returns,
         risk_factors['HML'],
-        rolling_window=rolling_beta_window)
+        window=rolling_beta_window)
     rolling_beta_UMD = timeseries.rolling_beta(
         returns,
         risk_factors['UMD'],
-        rolling_window=rolling_beta_window)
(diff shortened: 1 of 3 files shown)


humdings:
I have some questions/concerns about our regression calculations before I do more work on this.

1. I am thinking we should return fitted ols objects rather than just the parameters everywhere. There's a lot of information we are throwing out by returning the parameters alone. Maybe we can break regression into a single function that is called by all the others to make sure we alway have access to the original object. Does anybody have an opinion on this.

2. We seen to use/return an intercept in some regression cases but not others. Are there any cases where we do not want to use intercepts to perform a regression?

3. I think 2 has caused some test cases to break or be written incorrectly, I plan to fix them up here as well.

twiecki:
@humdings I agree with all your points. My only concern regarding returning 1. is usability and principle of least surprise (and returning an object rarely is not surprising). Maybe there's a way to have both though but I can't think of something I like..

gusgordon:
Are we going to get this in for the release?

twiecki:
I think this one is optional.

humdings:
bump!

Bringing this back up. I think we can include this in the initial ship. It's a pretty simple change that
just standardizes the functions used to perform regressions.

I removed the kwarg to return the model from each of the functions, but I do think having functions
available that do return the model are important. 

This still requires test rewrites which is why travis fails.

gusgordon:
OK, a lot of flake8 errors here have been fixed, and there are going to be merge conflicts here. Maybe it makes sense to reset this branch and just make the changes necessary from there? Let me know if you want me to do anything, too.

twiecki:
@gusgordon That's probably the best way forward.

MatthewGilbert:
Just thought I would add this in https://github.com/pydata/pandas/issues/6077. At some point pandas hopes to move the stats stuff to statsmodels and deprecate it, although as of now I don't believe there is any fast rolling or expanding OLS stuff in statsmodels.

twiecki:
Thanks @MatthewGilbert, good to know.

twiecki:
Closing due to inactivity. Hopefully we can implement this at a later point.

Issue #65, opened 2015-07-20
annual_returns incorrect in calendar mode


ACampbell22:
```
if style == 'calendar':
        num_years = len(returns) / 252
        df_cum_rets = cum_returns(returns, starting_value=100)
        start_value = df_cum_rets[0]
        end_value = df_cum_rets[-1]
        return ((end_value - start_value) / start_value) / num_years
```

Because the 252 is an integer, this function will return an inf for all return series shorter than one year.

Just gotta change the 252 to a float.

ssanderson:
This isn't actually true.  On line https://github.com/quantopian/pyfolio/pull/66/files#diff-dd656075c4a67ac1cd34b18065e18791L15 we do a `from __future__ import division`, which gives integer division the same semantics as floating point division.

twiecki:
I think this happened semi-simultaneously.

PR #66, opened 2015-07-21
changed 252 to 252.0 so calendar returns are correct

ACampbell22:


ssanderson:
Is the issue here that this is doing truncating integer division?  If so the right fix is to put `from __future__ import division` at the top of the file.

https://www.python.org/dev/peps/pep-0238/

justinlent:
I closed this issue which was original #65 with this commit last night:  https://github.com/quantopian/pyfolio/commit/7f585f9cca08ade116c851804c6da592af8da064

ssanderson:
@justinlent now that this is a public repo, we should *really* try to get out of the habit of pushing behavior changes directly to master.

Also, it looks like there's already a `__future__` division import in this file, which means the division here should have the exact same result regardless of the denominator here.  Can you guys show an example input where this is producing different results?

justinlent:
all good. feel free to revert my commit and i'll work on a branch in the future

ssanderson:
It's not worth changing at this point, but I'm still curious about why we think this changed any behavior.  As far as I can tell the change here is a no-op.

ssanderson:
Consider, for example:
```
In [1]: from __future__ import division

In [2]: 5 / 3
Out[2]: 1.6666666666666667

In [3]: 5 / 3.0
Out[3]: 1.6666666666666667
```

twiecki:
Agree with @ssanderson. https://github.com/quantopian/pyfolio/commit/38198f934ddfe9b74bae8506fb0a33d982707e89 implemented the right way to address this issue.

PR #74, opened 2015-07-21
ENH: Now downloads and stores factors in HDF5 from Dartmouth. Closes #44.

### LINKED ISSUE #44: pull Fama-French risk factors directly from the academic website
We should look into directly pulling this data since the csv can get out of date really quickly.  Maybe we supply a default csv (if we can) as well as expose a function to pull the data from the university page

We also might want to confirm if there are any licenses or issues with redistributing it. @twiecki maybe you have some experience understanding redistribution of data in the OSS world?

gusgordon:


gusgordon:
No module named requests. How should I fix this?

pyfolio/utils.py
```diff
@@ -126,25 +130,38 @@ def load_portfolio_risk_factors(filepath_prefix=None):
...
-        time.mktime(time.strptime(str(t), "%Y%m%d"))) for t in factors.index]
-
-    five_factors = factors.join(mom)
-    # transform the returns from percent space to raw values (to be consistent
-    # with our portoflio returns values)
-    five_factors = five_factors / 100
+    five_factors = None
+
+    # If it's been more than two days since we updated, redownload CSVs
+    if time.time() - os.path.getmtime('data/factors.h5') > 60*60*24*2:
```
twiecki: I think this could be done more elegantly with a pandas timedelta: http://pandas.pydata.org/pandas-docs/stable/timeseries.html#period

gusgordon:
Does this look good to merge? @twiecki

pyfolio/utils.py
```diff
@@ -126,25 +130,39 @@ def load_portfolio_risk_factors(filepath_prefix=None):
...
+            factors_csv = factors_zip.read('F-F_Research_Data_'
+                                           'Factors_daily.CSV')
+            factors_csv = factors_csv.split('\r\n\r\n')[1]
+
+            factors = pd.DataFrame.from_csv(StringIO(factors_csv), sep=',')
+            umd = pd.DataFrame.from_csv(StringIO(umd_csv), sep=',')
+
+            five_factors = factors.join(umd)
+            five_factors = five_factors / 100
+            five_factors.to_hdf('data/factors.h5', 'df')
```
twiecki: I think this might only work if you're importing pyfolio from the source dir. Might have to do:
```python
-    if filepath_prefix is None:
-        import pyfolio
-        filepath = os.path.join(os.path.dirname(pyfolio.__file__), 'data')
-    else:
-        filepath = filepath_prefix
```
But you can try.

Issue #84, opened 2015-07-23
Add more detailed descriptions of certain variables to docstrings


gusgordon:
As scotty pointed out. I think this will just be needed for the pandas objects that are passed.

PR #86, opened 2015-07-24
Deprecate backtest_pct, optionalize live_start_date, cache SPY, remove backtest2

### LINKED ISSUE #81: Remove live_algo_start date and replace with returns_out_of_sample

gusgordon:
A bunch of my fixes that we discussed yesterday. Closes #81 (see details there) #82 and #84. Also updates example NBs.

Not sure if the way I'm saving/loading files with these try/excepts is good. I think it should work ok, I tried to pay attention to cases when there wouldn't be write permission but not sure if that matters.

@twiecki @ssanderson @humdings please take a look if you get the chance.


twiecki:
Looks great!

PR #87, opened 2015-07-25
[MAINT] Remove extract_rets_pos_txn_from_backtest_obj

humdings:
The code to extract positions and transaction code from
the research env BacktestResult object has been moved
to qexec where it can be managed internally.

If qexec code changes extract_rets_pos_txn_from_backtest_obj
could break so removing it from pyfolio seems like a good call.


twiecki:
```
0.50s$ flake8 pyfolio
pyfolio/utils.py:118:80: E501 line too long (80 > 79 characters)
```

twiecki:
:+1:

PR #88, opened 2015-07-27
Make application of plotting style optional and only set during plotting

twiecki:
This also changes the plot style a bit.

Looking for feedback, not sure this is the best way to go about this.


gusgordon:
Looks good to me.

pyfolio/plotting.py
```diff
@@ -30,18 +30,21 @@
...
+    if rc is None:
+        rc = {}
+
+    rc_default = {'lines.linewidth': 1.5,
+                  'axes.facecolor': '0.995',
+                  'figure.facecolor': '0.97'}
+
+    # Add defaults if they do not exist
+    for name, val in rc_default.iteritems():
+        if name not in rc:
```
ssanderson: `rc.setdefault(name, val)`

pyfolio/plotting.py
```diff
@@ -30,18 +30,21 @@
...
+def set_context(context='notebook', font_scale=1.5, rc=None):
+    if rc is None:
+        rc = {}
+
+    rc_default = {'lines.linewidth': 1.5,
+                  'axes.facecolor': '0.995',
+                  'figure.facecolor': '0.97'}
+
+    # Add defaults if they do not exist
+    for name, val in rc_default.iteritems():
```
ssanderson: This is broken on Py3.

pyfolio/plotting.py
```diff
@@ -30,18 +30,21 @@
 from . import pos
 
 
-def set_plot_defaults():
-    """
-    Sets default plotting/charting colors/styles.
-    """
+def set_context(context='notebook', font_scale=1.5, rc=None):
```
ssanderson: IMO `set_context` is a misleading name for something that's still being used as a context manager.  I would just call this `context` and use it as:

```
with plotting.context(...):
    do stuff
```

pyfolio/tears.py
```diff
@@ -64,8 +64,14 @@ def create_returns_tear_sheet(returns, live_start_date=None,
         Daily non-cumulative returns of the first benchmark.
     return_fig : boolean, optional
         If True, returns the figure that was plotted on.
+    set_context : boolean, optional
+        If True, set default plotting style context.
     """
 
+    if set_context:
```
ssanderson: Is there a reason we're not using a proper `with` block here?  Elsewhere I've used the following idiom for conditional contextmanagers:

```
from contextlib import contextmanager

@contextmanager
def nullctx():
    "No-op contextmanager"
    yield


# Inside some function that maybe wants to use a context manager:
with some_context_manager if some_expression else nullctx():
    # Do context things.
```

pyfolio/tears.py
```diff
@@ -347,8 +384,14 @@ def create_bayesian_tear_sheet(returns, benchmark_rets, live_start_date,
         trading, after its backtest period.
     return_fig : boolean, optional
         If True, returns the figure that was plotted on.
+    set_context : boolean, optional
+        If True, set default plotting style context.
     """
 
+    if set_context:
```
ssanderson: If this boilerplate is going to be everywhere, we might want to factor it out into a decorator.

pyfolio/plotting.py
twiecki: neat!

pyfolio/tears.py
richafrank: I've found ExitStack also a good tool for conditional context managers.

ssanderson: @richafrank I think ExitStack requires contextlib2 as a dependency on Py2?  We may already have that though given that we've got a lot of heavyweight SciPy modules as dependencies.

richafrank: That's right @ssanderson.  I couldn't say whether it's already required.

twiecki:
@ssanderson I love all of these, will push an update soon.

twiecki:
@ssanderson Incorporated your feedback. I removed the noop context manager and replaced it with an `if` which I think is favorable due do reduced complexity and dependency.

pyfolio/plotting.py
```diff
@@ -30,18 +30,32 @@
 from . import pos
 
 
-def set_plot_defaults():
-    """
-    Sets default plotting/charting colors/styles.
-    """
+def plotting_context(func):
+    """Decorator to set plotting context during function call."""
+    def call_w_context(*args, **kwargs):
```
ssanderson: This should be decorated with  `functools.wraps(func)` so we don't lose docstrings and method signatures when inspecting interactively.

pyfolio/plotting.py
```diff
@@ -30,18 +30,32 @@
...
+        set_context = kwargs.pop('set_context', True)
+        if set_context:
+            with context():
+                return func(*args, **kwargs)
+        else:
+            return func(*args, **kwargs)
+    return call_w_context
+
+
+def context(context='notebook', font_scale=1.5, rc=None):
```
ssanderson: Is there value in having this accept arguments if it's only called in one place with all defaults?  Do we expect to users to use this standalone somehow?  If so, we should document that expectation.

ssanderson: One way to make this useful would be to replace the `set_context` boolean arg below with a `plotting_context` arg, which will be used if set, and will use the defaults here if `None`.  Then users could construct their preferred plotting context with this function and pass it to the plotting functions.

twiecki: Yeah, good point. I put that in for future use but there's currently no path to it. Your second idea is interesting but I'm worried that it breaks with the suggested way that these style context managers usually work. If I'm not mistaken, a user couldn't do `with set_style('my_style'): create_tear(set_style=False)` anymore but would have to pass the context in.

twiecki: I suppose we could check for `False` but it starts to feel like too many ways to do one thing.

twiecki: Actually it is useful already, a user could do:
```python
with pf.plotting.context(font_scale=2):
    pf.create_tear_sheet(set_style=False)
```

I think that's a good way to use it. Will add docs.

ssanderson: I think I still have a slight preference for
```
my_context = pf.plotting.context(...)
pf.create_tear_sheet(context=my_context)
```
I don't have a super strong preference either way though.  Thoughts @gusgordon @humdings @justinlent ?

twiecki: The other issue I think would be that we would always override a user-set style (e.g. via matplotlibrc file) as you either pass in your own, or use the default one, but you can't disable.

ssanderson: Fair enough.  What you've got proposed above makes sense to me then.

twiecki:
Thanks for the review @ssanderson, @gusgordon and @richafrank.

PR #91, opened 2015-07-27
MAINT: Now emits warnings about being unable to load/save HDF files. Closes #90.

### LINKED ISSUE #90: get_symbol_rets should warn about error
I couldn't get caching because of:
ImportError: HDFStore requires PyTables, "libhdf5.so.7: cannot open shared object file: No such file or directory" problem importing

But since we suppress the exception in get_symbol_rets this never gets displayed.

gusgordon:
@twiecki I am not sure about this - I am able to save/load the SPY data fine, but the factors data keeps spitting the same warning. I don't see what I'm doing differently, but I may be missing something, or it may be some problem with my environment.

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -20,6 +20,7 @@ from os.path import (
     getmtime,
     join,
 )
+import warnings
 
 from datetime import datetime
 
@@ -117,8 +118,9 @@ def get_symbol_rets(symbol):
             if datetime.now() - pd.to_datetime(
                     getmtime(filepath), unit='s') < pd.Timedelta(days=1):
                 rets = pd.read_hdf(filepath, 'df')
-        except:
-            pass
+        except Exception as e:
+            warnings.warn('Failed to load SPY from local, ' +
+                          'downloading from Yahoo instead: %s' % e)
 
     if rets is None:
         px = web.get_data_yahoo(symbol, start='1/1/1970')
@@ -129,8 +131,8 @@ def get_symbol_rets(symbol):
         if symbol == 'SPY':
             try:
                 rets.to_hdf(filepath, 'df')
-            except:
-                pass
+            except Exception as e:
+                warnings.warn('Failed to save SPY locally: %s' % e)
 
     return rets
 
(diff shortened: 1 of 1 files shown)


pyfolio/utils.py
```diff
@@ -117,8 +118,9 @@ def get_symbol_rets(symbol):
             if datetime.now() - pd.to_datetime(
                     getmtime(filepath), unit='s') < pd.Timedelta(days=1):
                 rets = pd.read_hdf(filepath, 'df')
-        except:
-            pass
+        except Exception as e:
```
ssanderson: Why are we swallowing the exception here?  Can we usefully recover from returning `None` as a benchmark?  If not, this is just going to create a confusing exception somewhere else that's further away from the source of the real problem.

pyfolio/utils.py
```diff
@@ -129,8 +131,8 @@ def get_symbol_rets(symbol):
         if symbol == 'SPY':
             try:
                 rets.to_hdf(filepath, 'df')
-            except:
-                pass
+            except Exception as e:
+                warnings.warn('Failed to save SPY locally: %s' % e)
```
ssanderson: This should print the path it tried to save to, since the most likely cause of the error is that the path doesn't exist or doesn't have appropriate permissions, both of which can be diagnosed easily with access to the target path.

pyfolio/utils.py
gusgordon: Sure. I'll also add some way to force-download the benchmark in case of exception.

pyfolio/utils.py
gusgordon: Should I still be printing the exception, or just the path?

pyfolio/utils.py
ssanderson: Also, we shouldn't be swallowing **any** exception here.  There are lots of exceptions that we don't want to catch (e.g. `MemoryError`, `KeyboardInterrupt`).  If there are specific exceptions we expect to be thrown here, we should catch them explicitly.

gusgordon: OK. Actually, why is it of concern for us to automatically download from Yahoo if we already do that sometimes anyway? Can we not just throw the (correct) exception, and download the data? I don't think there is a reason to be concerned as long as we aren't doing this behind users' backs. Maybe we can have some message notifying users when downloads occur. Then, if they are concerned with using outside data, they can pipe in their own benchmark or F-F factors. I imagine any firm/party wanting to use pyfolio, and concerned about using external data, would first familiarize themselves with it to the extent to where they would see messages about the downloads if we show them. Thoughts?

gusgordon:
I'm gonna wait until https://github.com/quantopian/pyfolio/pull/92 is done to get back at this.

gusgordon:
Overridden by 3c651a372338f99fa06531d0ab650bb04a9fc18f

PR #92, opened 2015-07-27
[ENH] Add returns function registration.

humdings:
Here is a really simple implementation of a returns function
registry mechanism.

This doesn't feel like a great way to go about this but it works.
Maybe we can use a class/decorator pattern but I'm not sure what
schools of thought are out there on this sort of thing.

Any thoughts on this, if it's fine I'll document it up and submit a PR.

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -30,6 +30,7 @@ import pandas.io.data as web
 
 import zipfile
 
+
 try:
     # For Python 3.0 and later
     from urllib.request import urlopen
@@ -92,7 +93,7 @@ def round_two_dec_places(x):
     return np.round(x, 2)
 
 
-def get_symbol_rets(symbol):
+def default_returns_func(symbol):
     """
     Gets returns for a symbol.
 
@@ -259,3 +260,16 @@ def extract_rets_pos_txn_from_zipline(backtest):
     transactions.index.tz = None
 
     return returns, positions, transactions, gross_lev
+
+
+SETTINGS = {
+    'returns_func': default_returns_func
+}
+
+def register_return_func(func):
+    SETTINGS['returns_func'] = func
+
+
+def get_symbol_rets(symbol):
+    return SETTINGS['returns_func'](symbol)
+
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -30,6 +30,8 @@ import pandas.io.data as web
 
 import zipfile
 
+import functools
+
 
 try:
     # For Python 3.0 and later
@@ -93,17 +95,18 @@ def round_two_dec_places(x):
     return np.round(x, 2)
 
 
-def default_returns_func(symbol):
+def default_returns_func(symbol, default_kwargs=None):
     """
     Gets returns for a symbol.
-
     Queries Yahoo Finance. Attempts to cache SPY in HDF5.
-
     Parameters
     ----------
     symbol : str
         Ticker symbol, e.g. APPL.
-
+    default_kwargs : dict
+        currently ignored but exists for
+        compatability with the required
+        returns_func signature.
     Returns
     -------
     pd.DataFrame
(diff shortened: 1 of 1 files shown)


twiecki:
Short and sweet, I think this is fine but other folks might have more ideas. Definitely needs some docs.

twiecki:
@ssanderson would be a good person to take a look I think.

pyfolio/utils.py
```diff
@@ -259,3 +260,45 @@ def extract_rets_pos_txn_from_zipline(backtest):
...
+        returns a pandas.Series of asset returns.
+
+    Returns
+    -------
+    None
+    """
+    SETTINGS['returns_func'] = func
+
+
+def get_symbol_rets(symbol):
```
twiecki: Probably want to pass *args and **kwargs.

ssanderson: @twiecki I think the signature is correct as written.  We know how this is going to be called by PyFolio internally; the function registered by the user should have the same signature as our default implementation.  What use-case do you imagine supporting by accepting arbitrary inputs?

humdings: I was initially thinking `*args, **kwargs` would make sense. A use case I can think of is somebody querying an internal database, they might need to pass db/table names or dates. What if we made the signature 
`get_symbol_rets(symbol, args=None, kwargs=None)`
Then the user would be expected to pass a tuple and/or dict of their required parameters. 

I was also planning on adding start/end dates to this, I'm sure data since 1970 isn't always required.

twiecki: Yeah, that's my thinking. While theoretically it'd be nice to set an API explicitly I think this will clash with reality where people might have to pass in certain arguments (credentials are a good example).

humdings: @twiecki @ssanderson What's the vote on this for the initial shipping? I am thinking
`get_symbol_rets(symbol, kwargs=None)` is sufficient without the positional args tuple.

humdings:
@ssanderson @twiecki On second thought I think we should not add the additional arguments to get_symbol_rets at this time.

It will require changing a few tear sheet function calls where the benchmark series is not provided. I think we can leave this as is to not risk breaking anything before the initial release. We can revisit this when we have more testing in place.

twiecki:
Why would it require changing tear sheet functions?

humdings:
Each of the tear sheet functions has a block like this
```
if benchmark_rets is None:
        benchmark_rets = utils.get_symbol_rets('SPY')
```
if we add a kwargs argument to `get_symbol_rets`, we will have to add it here as well.

pyfolio/utils.py
ssanderson: @humdings @twiecki how would you thread those additional arguments to `get_symbol_rets`.  The whole point of this function is that it's being called behind the scenes by `create_*_tearsheet` with a symbol.  If you wanted to forward arguments, then all the tearsheet functions would need to accept additional args and kwargs, at which point you might as well just load the benchmark explicitly as an array and pass it in.

If your symbol returns function needs database credentials, it can close over them like this:

```
creds = get_db_credentials()
connection = DatabaseConnection(creds)

def my_benchmark_func(symbol):
    return connection.make_some_query(...)
```

This is also a perfectly good use-case for `functools.partial`.

humdings: I realized the same thing a bit ago (see https://github.com/quantopian/pyfolio/pull/92#issuecomment-125677310) I think `functools.partial` will work pretty well to avoid the extra parameters all over. How about one of the methods below? I'm think the second has less potential issues in the future. 

```
def register_return_func(func, default_kwargs):
    SETTINGS['returns_func'] = functools.partial(func, **default_kwargs)

def register_return_func(func, default_kwargs):
    SETTINGS['returns_func'] = functools.partial(func, default_kwargs=default_kwargs)
```

ssanderson: @humdings if the user wants supply a partial, I don't think it's that hard for them to do it themselves.  I don't think it's worth providing an API that's just wrapping trivial standard library code.

twiecki: OK, `partial` is a pretty good alternative. Lets stick with the explicit API then.

humdings:
I was happy with the use of `functools.partial` to avoid passing extra parameters around so I pushed that change. I'm satisfied with this branch if anybody would like to give it a look over.

pyfolio/utils.py
humdings: Damn. I pushed with a partial before reading these last couple comments. Should I roll back to the single 'symbol' argument function and call it a day?

twiecki: We can probably merge it then, yeah.

ssanderson: @twiecki do you mean merge with or without the default_kwargs?  My vote would be (fairly strongly) in favor of just accepting a callable.

humdings:
Yes @twiecki I think this can be merged as is. I'm fine with the partial function, I don't think it makes anything less useable, unless @ssanderson thinks we should roll back to the single argument and leave using a partial to the user I say go ahead and merge.

pyfolio/utils.py
twiecki: @ssanderson Agreed, I meant without the default_kwargs.

humdings: Sounds good, I'll remove default_kwargs now then.

humdings:
We're good to go on merging this.

twiecki:
:+1: Thanks, @humdings (and @ssanderson for review)!

PR #93, opened 2015-07-27
[TST] Switched tests to use AssertAlmostEqual

humdings:
A few of the tests on my local machine have been consistently
been failing due to rounding errors so I changed them to use
AssertAlmostEqual. I have them checking out to 8 decimal places.

I'm pretty sure that other tests will need this change as well.


twiecki:
Yep, that makes sense to me.

PR #94, opened 2015-07-28
[BUG] Ensure utc timezones

humdings:
In research there was an issue converting timestamps to UTC, this makes sure the data returned by `get_symbol_rets` is tz-aware with a UTC timezone. It also added a method `get_utc_date` which returns the UTC representation of a timestamp. If the date is not tz-aware, it is given a UTC timezone, if it is,
it is simply converted to UTC.

I did not have the dates normalized to midnight so that doing intraday analysis is still possible in the future.

Hopefully this fixes the issues in research!

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -38,7 +38,7 @@ import matplotlib.pyplot as plt
 import matplotlib.gridspec as gridspec
 import seaborn as sns
 
-from . utils import normalize_date
+from . utils import get_utc_date
 
 
 @plotting_context
@@ -78,8 +78,8 @@ def create_returns_tear_sheet(returns, live_start_date=None,
         benchmark_rets = utils.get_symbol_rets('SPY')
         # If the strategy's history is longer than the benchmark's, limit
         # strategy
-        benchmark_start = normalize_date(benchmark_rets.index[0])
-        returns_start = normalize_date(returns.index[0])
+        benchmark_start = get_utc_date(benchmark_rets.index[0])
+        returns_start = get_utc_date(returns.index[0])
         if returns_start < benchmark_start:
             returns = returns[returns.index > benchmark_rets.index[0]]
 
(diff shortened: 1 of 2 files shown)


pyfolio/tears.py
```diff
@@ -76,8 +77,8 @@ def create_returns_tear_sheet(returns, live_start_date=None,
         benchmark_rets = utils.get_symbol_rets('SPY')
         # If the strategy's history is longer than the benchmark's, limit
         # strategy
-        benchmark_start = benchmark_rets.index[0].astimezone(pytz.UTC)
-        returns_start = returns.index[0].astimezone(pytz.UTC)
+        benchmark_start = get_utc_date(benchmark_rets.index[0])
```
twiecki: Why are we only converting the start and end dates rather then the whole index?

humdings: I'm not sure, I did not alter the original functionality, just ensured time zone conversion would be robust. 

My hunch is that we should do input checking and ensure UTC time stamps for the indexes.

twiecki: I wonder if we should just pass in the right thing from research. Either the returns or the benchmark must be tz-naive.

For general input checking I think it'd be great to use https://github.com/TomAugspurger/engarde.

humdings: In research all of the pricing data is UTC timestamped, I'd vote to have pyfolio do the same. The data returned by yahoo was tz-naive but this PR converts it to UTC before it's returned. I think this will fix the error we saw on staging. 

I'm also not a fan of the .astimezone() method because it's not a DatetimeIndex method. .tz_convert() and .tz_localize() work for both DatetimeIndex/Timestamp inputs.

twiecki: I see. So in that case do you think we should still convert the indices here?

humdings: I'll test it with the conversion removed and see what happens when tz-naive and aware returns are used and get back to you.

twiecki:
So this issue seemed to have gone away now? Or was it fixed by https://github.com/quantopian/pyfolio/pull/96?

humdings:
I'm pretty sure this was fixed by https://github.com/quantopian/pyfolio/pull/96 and by having the default returns function return a tz-aware series.

twiecki:
OK, closing this one then.

PR #96, opened 2015-07-29
[MAINT] Removed setting transaction timezones to None

humdings:
There was one line that was explicitly setting the time zone
of the transactions dataframe to None. By removing that line looks
like all of the data returned has UTC timezones.

I also set the backtest dataframe's timezone to UTC if it was None.

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -243,6 +243,8 @@ def extract_rets_pos_txn_from_zipline(backtest):
     """
 
     backtest.index = backtest.index.normalize()
+    if backtest.index.tzinfo is None:
+        backtest.index = backtest.index.tz_localize('UTC')
     returns = backtest.returns
     gross_lev = backtest.gross_leverage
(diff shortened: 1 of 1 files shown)


pyfolio/utils.py
```diff
@@ -243,6 +243,8 @@ def extract_rets_pos_txn_from_zipline(backtest):
     """
 
     backtest.index = backtest.index.normalize()
+    if backtest.index.tzinfo is None:
+        backtest.index = backtest.index.tz_localize('UTC')
```
gusgordon: Do we want to just use the get_utc_date function you wrote in #94?

twiecki:
:+1:

pyfolio/utils.py
humdings: In this case .tz_localize is fine because tzinfo has to be None. tz_localize works for tz-naive timestamps, and .tz_convert only works for tz-aware ones.

PR #97, opened 2015-07-29
[BUG] KeyError Fix in pos.get_top_long_short_abs

humdings:
Using `series[:N]` in research is causing a KeyError. To fix this
I cleaned up the function by using `series.nlargest(N)`. This
also allowed me to delete the sorting methods from the function.


twiecki:
:+1:

Issue #95, opened 2015-07-29
get_symbol_rets should accept start and end times and only update cache if necessary


twiecki:
Here's an idea:

@humdings already made the suggestion of adding start and end dates as keyword arguments to `get_symbol_rets` which I think is a good idea. Then once we have that and SPY gets requested, we could check if the end_date is later than what we have cached and only update the cache if necessary. Now we are updating if it's older than a day which is unnecessary and causes daily re-downloading of SPY which can take a while.

Thoughts?

gusgordon:
Yeah, good idea, this is the correct way to handle it.

To tack on here, @ssanderson and I talked a little more about whether or not to force users to have to explicitly call the benchmark. So for example, it would be:
`create_full_tear_sheet(returns, benchmark_returns=get_symbol_rets('SPY'))`

I don't think this is too harmful to ease of use, but IMO still not worth it. I think the best solution is to have a warning or something whenever we are downloading data. See https://github.com/quantopian/pyfolio/pull/91. There are definitely a few nontrivial things we should do with this function.

PR #99, opened 2015-07-31
Refactor get symbol rets

twiecki:
Add start and end date kws to get_symbol_rets and refactor caching logic to only update cache if necessary.

Closes https://github.com/quantopian/pyfolio/issues/95 and https://github.com/quantopian/pyfolio/issues/90

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -25,7 +25,6 @@ from datetime import datetime
 
 import pandas as pd
 import numpy as np
-import zlib
 import pandas.io.data as web
 
 import zipfile
@@ -49,24 +48,6 @@ def pyfolio_root():
 
 def data_path(name):
     return join(pyfolio_root(), 'data', name)
-
-
-def json_to_obj(json):
-    """
-    Converts a JSON string to a DataFrame.
-
-    Parameters
-    ----------
-    json : str
-        Data to convert.
-
-    Returns
-    -------
-    pd.DataFrame
-        The converted data.
-    """
-
-    return pd.json.loads(str(zlib.decompress(json)))
 
 
 def one_dec_places(x, pos):
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -20,6 +20,7 @@ from os.path import (
     getmtime,
     join,
 )
+import warnings
 
 from datetime import datetime
 
@@ -116,7 +117,13 @@ def default_returns_func(symbol, start=None, end=None):
             # Download most-recent SPY to update cache
             rets = get_symbol_from_yahoo(symbol, start='1/1/1970',
                                          end=datetime.now())
-            rets.to_hdf(filepath, 'df')
+            try:
+                rets.to_hdf(filepath, 'df')
+            except IOError as e:
+                warnings.warn('Could not update cache {}.'
+                              'Exception: {}'.format(filepath, e),
+                              UserWarning)
+
         rets = rets[start:end]
     else:
         rets = get_symbol_from_yahoo(symbol, start=start, end=end)
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -74,7 +74,7 @@ def round_two_dec_places(x):
     return np.round(x, 2)
 
 
-def default_returns_func(symbol):
+def default_returns_func(symbol, start=None, end=None):
     """
     Gets returns for a symbol.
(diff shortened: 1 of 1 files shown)


twiecki:
@humdings can you help review?

pyfolio/utils.py
```diff
@@ -108,29 +95,31 @@ def default_returns_func(symbol):
...
+    if symbol == 'SPY':
+        filepath = data_path('spy.h5')
+        # Is cache recent enough?
+        if pd.to_datetime(getmtime(filepath), unit='s') >= end:
+            rets = pd.read_hdf(filepath, 'df')
+        else:
+            # Download most-recent SPY to update cache
+            rets = get_symbol_from_yahoo(symbol, start='1/1/1970',
+                                         end=datetime.now())
+            rets.to_hdf(filepath, 'df')
```
gusgordon: I'm fairly certain that this will throw an exception in the case when the directory is write-protected. This was giving me some trouble before.

twiecki: Good point, I suppose if the caching isn't working there's no way to get returns. So maybe we turn the exception into a warning and just don't update the cache.

pyfolio/utils.py
```diff
@@ -108,29 +95,31 @@ def default_returns_func(symbol):
         Daily returns for the symbol.
          - See full explanation in tears.create_full_tear_sheet (returns).
     """
+    start = '1/1/1970' if start is None else start
+    end = datetime.now() if end is None else end
```
gusgordon: I think we can do
```
if start is None: start = '1/1/1970'
if end is None: end = datetime.now()
```

pyfolio/utils.py
```diff
@@ -108,29 +96,37 @@ def default_returns_func(symbol):
...
```
humdings: The pandas already defaults to today in `web.get_data_yahoo` so this isn't necessary except for readability purposes.

gusgordon:
Do you think it's worth it to alert people when we are downloading data from Yahoo when they don't explicitly call the Yahoo data fetcher?

pyfolio/utils.py
twiecki: Actually we do need these set for below we do `rets = rets[start:end]`.

twiecki:
@gusgordon I think the fact that we are downloading data from yahoo shouldn't really be surprising to users. Plus, it's already in the doc string. Or maybe I'm misunderstanding?

gusgordon:
OK, yeah, I agree. Anyone who is concerned with downloading/using data from Yahoo should be bent on looking at the docs.

twiecki:
@gusgordon Or do you mean for when we automatically pull the benchmark? Actually you might be right we want to display something if we're downloading that just because it takes a while.

gusgordon:
Yeah, just for when we automatically get the benchmark (when the user doesn't him/herself explicitly make some call to get external data). That's true too. I'm not convinced either way.

pyfolio/utils.py
```diff
@@ -108,29 +96,40 @@ def default_returns_func(symbol):
         Daily returns for the symbol.
          - See full explanation in tears.create_full_tear_sheet (returns).
     """
+    if start is None:
+        start = '1/1/1970'
+    if end is None:
+        end = datetime.now()
```
humdings: I don't think we should use `datetime.now()` because we only need daily granularity. 
We might want to use `end = pd.Timestamp.now().date()`. 

`pd.io.data.get_data_yahoo` does not return the current trading days data while the market is open, and I'm not sure how long after market close they add the most recent day's data so we can probably leave a one day lag before updating the file.

pyfolio/utils.py
```diff
@@ -108,29 +96,40 @@ def default_returns_func(symbol):
...
-                    getmtime(filepath), unit='s') < pd.Timedelta(days=1):
-                rets = pd.read_hdf(filepath, 'df')
-        except:
-            pass
+    start = pd.Timestamp(start)
+    end = pd.Timestamp(end)
 
-    if rets is None:
-        px = web.get_data_yahoo(symbol, start='1/1/1970')
+    def get_symbol_from_yahoo(symbol, start=None, end=None):
```
humdings: Is this nested because we don't wan't to expose the function? It seems pretty harmless and could be useful for people if we expose it.

humdings:
I think this is good as written. I have some small complaints but it will work fine as is. I will keep track of how often the h5 updates to see if we should use dates instead of .now.

PR #102, opened 2015-08-03
[BUG] Make Benchmark tz-aware

humdings:
This is a one line change that makes get_data_from_yahoo return
a timezone aware series (UTC).

Once the benchmark file gets overwritten the h5 file will have the
timezone so this should fix any benchmark timezone issues for
people using pyfolio locally.


twiecki:
That looks good. Maybe should update https://github.com/quantopian/pyfolio/blob/master/pyfolio/data/spy.h5 as well?

humdings:
Done, I pushed the tz-aware spy.h5 file

twiecki:
Did you also update the file to have more recent data (i.e. until yesterday)?

humdings:
Yes, I ran `get_symbol_rets('SPY')` and confirmed the data is up to 2015-07-31

twiecki:
:+1:

PR #112, opened 2015-08-04
[BUG] makes get_utc_timestamp work with datetime.datetime objs.

humdings:
Python datetimes were throwing an error in `get_utc_timestamp`,
this fixes that issue for me locally. I used `pd.to_datetime` to
ensure that pd.Timestamps are used. The method works for arrays
as well as individual datetimes.

The example NBs run without error for me locally with this change.


twiecki:
:+1:

PR #115, opened 2015-08-05
ENH: All functions using live_start_date now accept it as a string. Closes #114.

### LINKED ISSUE #114: Should be able to pass in a string for live_start_date
this use to work, then broke when improving the timezone stuff
Basically any valid/parsable time string should be valid.  any string that 'just works' when passed to pd.Timestamp() is probably a reasonable bar

image

gusgordon:
@twiecki @humdings review if you get the chance, thanks.


humdings:
Looks good to me, I thought that had been done already in another PR. Either way, looks good.

twiecki:
:+1:

PR #116, opened 2015-08-07
PERF speed up for get_txn_vol calculation

humdings:
Straight forward refactor of get_txn_vol. The lambda
statements made the function hard to read and much slower
so I got rid of them.

Speed stats:
old func: 1 loops, best of 3: 399 ms per loop
new func: 100 loops, best of 3: 10.2 ms per loop


twiecki:
Very clear and concise. :+1:

Issue #104, opened 2015-08-03
Inconsistent Turnover Calculations


humdings:
We have a turnover function in pos.py, but `plotting.plot_turnover()` does not use it. Also, the two calculations are done differently. 

```
# plotting.plot_turnover
df_turnover = transactions.txn_volume / positions.abs().sum(axis='columns')

# pos.turnover
turnover = transactions_df.apply(
        lambda z: z.apply(
            lambda r: abs(r))).resample(period, 'sum').sum(axis=1)
portfolio_value = backtest_data_df.portfolio_value.resample(period, 'mean')
turnoverpct = turnover / portfolio_value
```
These two calculations do not produce the same output

PR #117, opened 2015-08-07
BUG Fix turnover calculation

humdings:
This fixed the incorrect turnover calculation in
`plotting.plot_turnover`.

Now the plot calls `pos.get_turnover` rather than calculating
the values.

### DIFF
--- a/pyfolio/pos.py
+++ b/pyfolio/pos.py
@@ -135,29 +135,37 @@ def extract_pos(positions, cash):
     return values
 
 
-def turnover(transactions_df, backtest_data_df, period='M'):
+def get_turnover(transactions, positions, period=None):
     """
-    Calculates the percent absolute value portfolio turnover.
+    Portfolio Turnover Rate:
+
+    Average value of purchases and sales divided
+    by the average portfolio value for the period.
+
+    If no period is provided the period is one time step.
 
     Parameters
     ----------
     transactions_df : pd.DataFrame
-        Contains transactional data.
-    backtest_data_df : pd.DataFrame
-        Contains backtest data, like positions.
+        Contains transactions data.
+        - See full explanation in tears.create_full_tear_sheet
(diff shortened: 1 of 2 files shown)


pyfolio/pos.py
```diff
@@ -135,29 +135,37 @@ def extract_pos(positions, cash):
...
-    portfolio_value = backtest_data_df.portfolio_value.resample(period, 'mean')
-    turnoverpct = turnover / portfolio_value
-    turnoverpct = turnoverpct.fillna(0)
-    return turnoverpct
+    traded_value = transactions.txn_volume
+    portfolio_value = positions.sum(axis=1)
+    if period is not None:
+        traded_value = traded_value.resample(period, how='sum')
+        portfolio_value = portfolio_value.resample(period, how='mean')
+    turnover = traded_value / 2.0
```
twiecki: When using constants like here it's useful to add a comment on how it came about.

twiecki:
Looks great! While we're here, do you think we should add unit tests?

twiecki:
Related issue: https://github.com/quantopian/pyfolio/issues/104

pyfolio/pos.py
humdings: Okay, the doc string above says it uses the average of buys & sells, that's where the 2.0 comes from. 

A lot of definitions use min(buys, sells) / average portfolio value, but we only store the abs value of buys + sells so averaging makes sense.

humdings:
Yes, we should definitely have unit tests, I'll get to that a bit later today.

pyfolio/pos.py
twiecki: I agree. Then I'd add something like `# traded_value contains the summed value from buys and sells; this computes the average of both`.

twiecki:
Great. I'll merge this in now as it fixes a bug. We can add a unittest later.

PR #119, opened 2015-08-14
MAINT Change format of cached SPY and Fama-French factors to csv inst…

twiecki:
…ead of hdf. Also fixes some more UTC issues.


twiecki:
The reason I changed it is because we ran into this issue: https://github.com/pydata/pandas/issues/8165

Also removes pytables as a dependency.

gusgordon:
Interesting, looks great to me.

gusgordon:
I think this is made redundant by #120?

twiecki:
Thanks for reviewing @gusgordon! Yeah, that branch is based off of this one.

PR #120, opened 2015-08-14
Bayesian refactor + new model

twiecki:


pyfolio/bayesian.py
```diff
@@ -165,6 +166,169 @@ def model_returns_t(data, samples=500):
...
+
+    This model replicates the example used in:
+
+    Kruschke, John. (2012) Bayesian estimation supersedes the t
+    test. Journal of Experimental Psychology: General.
+
+    The original pymc2 implementation was written by Andrew Straw and
+    can be found here: https://github.com/strawlab/best
+
+    Ported to PyMC3 by Thomas Wiecki (c) 2015.
```
gusgordon: Is this the same as your function in PyMC3? Either way we should change this line to pyfolio.

gusgordon:
Looks good from the parts I understood. What's different about this new one?

pyfolio/tears.py
```diff
@@ -549,12 +560,23 @@ def create_bayesian_tear_sheet(returns, benchmark_rets, live_start_date,
                           horizontalalignment='right',
                           transform=ax_ret_pred_week.transAxes)
 
+    # Run alpha beta model
+    benchmark_rets = benchmark_rets.loc[df_train.index]
+    trace_alpha_beta = bayesian.run_model('alpha_beta', df_train,
+                                          bmark=benchmark_rets, samples=2000)
```
gusgordon: Should we allow other sample sizes to be passed when we create the tear sheet?

pyfolio/bayesian.py
twiecki: Good point, yeah.

twiecki:
OK @gusgordon, addressed the feedback. Merging in.

PR #121, opened 2015-08-14
TST: Added tests for get_turnover

humdings:
This added a simple test case for get_turnover.
1. No transactions -> no turnover
2. 100% value traded daily -> 50% daily turnover

This was repeated with a monthly freq, I used 20 days
so the monthly result should be the sum of the daily result.


gusgordon:
Great!

twiecki:
:+1:

PR #130, opened 2015-08-26
MAINT Break out rolling Fama-French calculation.

twiecki:


pyfolio/timeseries.py
```diff
@@ -584,6 +586,55 @@ def rolling_multifactor_beta(returns, df_multi_factor, rolling_window=63):
     return out
 
 
+def rolling_risk_factors(returns, risk_factors=None,
+                         rolling_beta_window=63 * 2):
```
CaptainKanuk: Where does 63 * 2 come from?

pyfolio/timeseries.py
```diff
@@ -584,6 +586,55 @@ def rolling_multifactor_beta(returns, df_multi_factor, rolling_window=63):
     return out
 
 
+def rolling_risk_factors(returns, risk_factors=None,
```
CaptainKanuk: I don't like calling this function "risk_factors" when it only computes Fama-French right now. We should have a function that can compute risk against any given factor, and this should be a specific case that calls that imho

CaptainKanuk:
Made some comments.

pyfolio/timeseries.py
twiecki: Good question -- @justinlent ?

twiecki:
@CaptainKanuk Refactored in response to comments.

pyfolio/timeseries.py
justinlent: Delaney,
The 63 * 2 number is a bit of an artifact from the original prototype
code.  It's just supposed to be reflective of a 6-month window of trading
days (e.g. 21 * 6). we should probably just swap in the value 126 which is
the approx # of trading days in a 6-month period.  I'm also totally open to
other more robust approaches.

pyfolio/plotting.py
```diff
@@ -94,10 +94,10 @@ def context(context='notebook', font_scale=1.5, rc=None):
                                 rc=rc)
 
 
-def plot_rolling_risk_factors(
+def plot_rolling_fama_french(
         returns,
         risk_factors=None,
-        rolling_beta_window=63 * 2,
+        rolling_window=63 * 2,
```
twiecki: How about we refactor this to:
* Define a global variable `BDAYS_MONTH = 21`
* The default will become `BDAYS_MONTH * 6`

pyfolio/plotting.py
```diff
@@ -94,10 +96,10 @@ def context(context='notebook', font_scale=1.5, rc=None):
                                 rc=rc)
 
 
-def plot_rolling_risk_factors(
+def plot_rolling_fama_french(
         returns,
         risk_factors=None,
-        rolling_beta_window=63 * 2,
+        rolling_window=BDAYS_PER_MONTH * 6,
```
CaptainKanuk: Why is this 6 months default when many others are 3?

pyfolio/plotting.py
```diff
@@ -29,6 +29,8 @@
 from . import timeseries
 from . import pos
 
+from .utils import BDAYS_PER_MONTH
```
CaptainKanuk: Can we call this APPROX_BDAYS_PER_MONTH?

Also, I understand from Justin that @twiecki thinks that BDAYS (vs TDAYS) is good, as that is the naming convention in pandas. However, doesn't pandas have full calendar functionality that we could use in this case? Why not pass rolling window as a pandas `datetime.timedelta` object with a given number of months or days and then let pandas or zipline tradingcalendar figure out the actual number of days? The use case would be something like
```python
plot_rolling_fama_french(returns, rolling_window=datetime.timedelta(months=3))
plot_rolling_fama_french(returns, rolling_window=datetime.timedelta(days=90))
```

pyfolio/timeseries.py
```diff
@@ -584,6 +596,40 @@ def rolling_multifactor_beta(returns, df_multi_factor, rolling_window=63):
     return out
 
 
+def rolling_fama_french(returns, risk_factors=None,
```
CaptainKanuk: Like this workflow a lot more, robust to people using their own factors now :+1:

pyfolio/timeseries.py
```diff
@@ -512,7 +516,8 @@ def calc_multifactor(returns, factors):
     return results.params
 
 
-def rolling_beta(returns, benchmark_rets, rolling_window=63):
+def rolling_beta(returns, benchmark_rets,
```
CaptainKanuk: I have a weird aversion to calling this `benchmark_rets`, because it's not always the benchmark to which I want to compute beta. I understand, though, that this is the 90% use case. Might it make any sense to rename to something like `factor_rets`, and then in the help string say

factor_rets : pd.Seires or pd.DataFrame
Daily noncumulative returns of the factor to which beta is computed. Usually a benchmark such as the market.

I'm not strong on this issue, more a point for discussion because I'm doing all this in the lectures and seeing how academics refer to these concepts.

pyfolio/plotting.py
CaptainKanuk: We could also overload the variable and do a typecheck in the function

WARNING: Syntax is approximate

```python
if type(rolling_window) is datetime.timedelta:
    num_days = rolling_window.days
elif type(rolling_window) is Number:
    num_days = rolling_window
else:
    raise TypeError

Comment (justinlent) [inline pyfolio/plotting.py]: @CaptainKanuk I like this timedelta approach, but I will let @twiecki weigh in. I'm just not sure if forcing a timedelta type of contruct (?) or pandas dependency may hamper potential usecases down the road that aren't pandas-friendly
ie: as we build out pyfolio to include additional analytical aspects, etc

File: pyfolio/timeseries.py (hunk 7 above)
Comment (humdings) [inline pyfolio/timeseries.py]: +1 on using `factor_rets` over `benchmark_rets`. It's more general and in line with academic/industry lingo.

Comment (twiecki) [inline pyfolio/timeseries.py]: Agreed. I changed it to `factor_returns` as the functions are unaware of which factor gets passed in. For that reason, I left `tears.py` to still use `benchmark_rets` as there is specific meaning associated. Make sense?

File: pyfolio/plotting.py (hunk 4 above)
Comment (twiecki) [inline pyfolio/plotting.py]: Yeah, I changed it to be 6 months throughout as I'm worried about stability of the estimates. 126 days should be pretty robust while 63 is certainly not enough to get reliable regression estimates. @justinlent what do you think?

File: pyfolio/plotting.py (hunk 5 above)
Comment (twiecki) [inline pyfolio/plotting.py]: I'm not huge on supporting `timedelta`. If someone really wants to use it, they can always call:
```python
plot_rolling_fama_french(returns, rolling_window=datetime.timedelta(months=3).days)
```

I'm open to renaming to `APPROX_BDAYS_PER_MONTH`.

twiecki:
OK, I think addressed most current comments.

pyfolio/timeseries.py
```diff
@@ -603,10 +611,11 @@ def calc_alpha_beta(returns, benchmark_rets):
...
-    """
+
+"""
 
     ret_index = returns.index
-    beta, alpha = sp.stats.linregress(benchmark_rets.loc[ret_index].values,
+    beta, alpha = sp.stats.linregress(factor_returns.loc[ret_index].values,
                                       returns.values)[:2]
 
     return alpha * 252, beta
```
CaptainKanuk: 252 should probably be a constant in utils, not hardcoded

pyfolio/plotting.py
```diff
@@ -692,10 +674,10 @@ def plot_rolling_beta(returns, benchmark_rets, rolling_beta_window=63,
     ax.set_title("Rolling Portfolio Beta to S&P 500")
     ax.set_ylabel('Beta')
     rb_1 = timeseries.rolling_beta(
-        returns, benchmark_rets, rolling_window=rolling_beta_window * 2)
+        returns, factor_returns, rolling_window=APPROX_BDAYS_PER_MONTH * 6)
     rb_1.plot(color='steelblue', lw=3, alpha=0.6, ax=ax, **kwargs)
     rb_2 = timeseries.rolling_beta(
-        returns, benchmark_rets, rolling_window=rolling_beta_window * 3)
+        returns, factor_returns, rolling_window=APPROX_BDAYS_PER_MONTH * 12)
```
CaptainKanuk: This function's docstring doesn't say why it plots two series.

CaptainKanuk:
A few comments, but looks good after those two.

Issue #124, opened 2015-08-19
Show '% of portfolio' traded instead of shares traded for turnover plots


justinlent:
where (% of portfolio) = ($ traded) / (portfolio value),

and,  ($ traded) = (stock price) * (# of shares)

![image](https://cloud.githubusercontent.com/assets/6541022/9371101/bd32189a-468a-11e5-8bde-d08a56c6b46f.png)

twiecki:
+1

humdings:
I think this is a pretty quick change, I agree that volume plots are pretty uninformative. 

The `plot_turnover` function is already plotted as a percentage of portfolio value, except it's 
`0.5*($ traded) / (portfolio value)`.

Maybe we can remove the volume plot from the tear sheet and replace the volume histogram with a histogram of the daily turnover rates. The turnover histogram would be a more informative plot.

twiecki:
+1

twiecki:
Closed with https://github.com/quantopian/pyfolio/pull/133. Thanks @humdings!

PR #133, opened 2015-09-02
MAINT replace volume hist plot with turnover hist plot

humdings:
This replaces the volume histogram with a histogram of the
daily turnover rate.

The volume plot is still in the txn_tearsheet because I did not want to change the
size of the grid in case it breaks something or looks funny.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -1053,6 +1053,39 @@ def plot_turnover(returns, transactions, positions,
     return ax
 
 
+def plot_daily_turnover_hist(transactions, positions,
+                             ax=None, **kwargs):
+    """Plots a histogram of daily turnover rates.
+
+    Parameters
+    ----------
+    transactions : pd.DataFrame
+        Daily transaction volume and dollar ammount.
+         - See full explanation in tears.create_full_tear_sheet.
+    positions : pd.DataFrame
+        Daily net position values.
+         - See full explanation in tears.create_full_tear_sheet.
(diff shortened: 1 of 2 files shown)


humdings:
This PR is addressing issue https://github.com/quantopian/pyfolio/issues/124

twiecki:
Will take a look soon.

pyfolio/plotting.py
```diff
@@ -1038,6 +1038,39 @@ def plot_turnover(returns, transactions, positions,
...
+    ax : matplotlib.Axes
+        The axes that were plotted on.
+
+    """
+
+    if ax is None:
+        ax = plt.gca()
+    turnover = pos.get_turnover(transactions, positions, period=None)
+    sns.distplot(turnover, ax=ax, **kwargs)
+    ax.set_title('Distribution of Daily Turnover Rates')
```
twiecki: Minor: I want to start refactoring this pattern to: `ax.set(title='...', xlabel='...')`.

twiecki:
LGTM. I think we can remove `plotting.plot_volume_per_day_hist`.

gusgordon:
Nice, good idea. I'm indifferent to removing `plotting.plot_volume_per_day_hist`. People still might want to use it standalone, but on the other hand if it's not worth showing in the tear sheets maybe it's not worth having at all.

twiecki:
Yeah I think we should be pretty eager to remove unused code to avoid bit
rot. It's easy to bring back if we do want it.

PR #139, opened 2015-09-12
Fix timestamps and add show()

### LINKED ISSUE #107: formatting fix: remove timestamps from peak/valley dates in drawdown table
image

gusgordon:
Closes #107.

Also added show() after the creation of each tear sheet so that they are displayed immediately upon completion. This changes the ordering of the data, because now the text going along with each tear sheet is displayed before its tear sheet, where as before all the text was at the beginning and all the tear sheets at the end. I think this makes more sense, even though it is a change. Let me know what you guys think @twiecki @humdings @justinlent


twiecki:
LGTM. Rebased and squashed some commits. Will merge once travis passes.

twiecki:
I definitely like how this groups the print() statements with the tear sheets. It makes me wonder if we shouldn't add some indicator that a new tear sheet is starting (but we can leave this for the future).

PR #150, opened 2015-09-25
[BUG] Fix #145 drawdown overlaps

### LINKED ISSUE #145: Overlapping drawdowns
The `get_top_drawdowns` function in `timeseries.py` sometimes returns overlapping drawdown periods, which should not happen. This code reproduces the issue:

```
import pyfolio as pf
stock_rets = pf.utils.get_symbol_rets('FB')
pf.timeseries.gen_drawdown_table(stock_rets)
```

which produces the following output:

screen shot 2015-09-22 at 14 24 48

Drawdown with index 6 starts on 2014-10-24, finishes on 2015-01-15
Drawdown with index 5 starts on 2014-10-28, which cannot be a peak as then it would have meant the end of the previous drawdown.

Problem occurs on `timeseries.py` lines 770-772:

```
            underwater = pd.concat(
                [underwater.loc[:peak].iloc[:-1],
                 underwater.loc[recovery:].iloc[1:]])
```

the actual peak is getting removed from the series, allowing a subsequent drawdown to engulf the one that just got cropped. The cropping should start on the index immediately after `peak`. Assuming that indexes are always day-based replacing line 771 by 

```
[underwater.loc[:peak + pd.to_timedelta("1d")].iloc[:-1]
```

seems to fix the bug.

jlopezpena:
Fix the drawdown overlaps described in #145, also making the `get_drawdown` function more efficient.

Add a regression test that checks that the recovery of each drawdown occurs before the peak of the next (when ordered by date). 

Note: The test requires being online to pull the FB stock data (or have it cached). I couldn't find any guidelines about tests. If that is not appropriate let me know and I will work on getting an offline version of it.

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -66,15 +66,15 @@ def create_full_tear_sheet(returns, positions=None, transactions=None,
         Daily net position values.
          - Time series of dollar amount invested in each position and cash.
          - Days where stocks are not held can be represented by 0 or NaN.
+         - Non-working capital is labelled 'cash'
          - Example:
             index         'AAPL'         'MSFT'          cash
             2004-01-09    13939.3800     -14012.9930     711.5585
             2004-01-12    14492.6300     -14624.8700     27.1821
             2004-01-13    -13853.2800    13653.6400      -43.6375
     transactions : pd.DataFrame, optional
-        Daily transaction volume and dollar ammount.
-        -  Time series of dollar amount of transactions per day,
-            and number of shares traded per day.
+        Daily transaction quantity (txn_shares) and dollar amount
+        (txn_volume).
         - Example:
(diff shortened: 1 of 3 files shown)


twiecki:
Awesome, thanks for pulling that together. The question re FB is a good one. Can you change it to SPY which we provide a cached version of?

jlopezpena:
Changed the test to use SPY, and lightly modified the check in order to get more informative messages in case of error (a test error should now show the dates for the overlapping drawdown periods rather than the uninformative `True does not equal False` that was showing before).

Reproducing the error for SPY required inspecting over a larger number of drawdowns (20 rather than 10)

Patch is also rebased on top of the latest master branch

twiecki:
LGTM, thanks!

PR #162, opened 2015-10-07
BUG Monthly returns distribution should not stack by year.

twiecki:
Out of monthly returns we created a year by month table, then dropped nans, then unstacked again to create the flat array for monthly returns distribution. Essentially throwing away data. 
Related: https://github.com/quantopian/qexec/issues/7890


twiecki:
Merged with https://github.com/quantopian/pyfolio/commit/2e8cbea4cadebdae05d61bfe57d457ad3d8cfc8d

PR #168, opened 2015-10-13
ENH Ability to make cones with multiple shades stdev regions

justinlent:


pyfolio/plotting.py
```diff
@@ -611,10 +611,18 @@ def plot_rolling_returns(
             label='Live', ax=ax, **kwargs)
 
         if cone_std is not None:
-            cone_df = timeseries.cone_rolling(
-                returns,
-                num_stdev=cone_std,
-                cone_fit_end_date=live_start_date)
+            # check to see if we're just rendering a single cone, or multiple
+            # cones at various stddevs, defined as elements of cone_std
+            if type(cone_std) is list:
```
twiecki: This should be `if isinstance(cone_std, list)`. It wouldn't allow passing of a tuple though.

pyfolio/plotting.py
```diff
@@ -641,7 +649,25 @@ def plot_rolling_returns(
             ax.fill_between(cone_df_live.index,
                             cone_df_live.sd_down,
                             cone_df_live.sd_up,
-                            color='red', alpha=0.30)
+                            color='steelblue', alpha=0.25)
+
+            if type(cone_std) is list:
```
twiecki: There might be a more elegant way to do this. I'll think abit on it.

justinlent: cool, yea the appropriate way to do type checking in Python seems to be a pretty controversial issue on stackoverflow :)  

**Sent from iPhone**




On Tue, Oct 13, 2015 at 6:51 AM -0700, "Thomas Wiecki" <notifications@github.com> wrote:












In pyfolio/plotting.py:
> @@ -641,7 +649,25 @@ def plot_rolling_returns(
>              ax.fill_between(cone_df_live.index,
>                              cone_df_live.sd_down,
>                              cone_df_live.sd_up,
> -                            color='red', alpha=0.30)
> +                            color='steelblue', alpha=0.25)
> +
> +            if type(cone_std) is list:



There might be a more elegant way to do this. I'll think abit on it.

justinlent: this is the best thread I found on type checking. actually maybe:   try/except TypeError:is the best preferred way to do thishttp://stackoverflow.com/questions/152580/whats-the-canonical-way-to-check-for-type-in-python

**Sent from iPhone**




On Tue, Oct 13, 2015 at 7:05 AM -0700, "Justin Lent" <justinlent@gmail.com> wrote:










cool, yea the appropriate way to do type checking in Python seems to be a pretty controversial issue on stackoverflow :)  

**Sent from iPhone**




On Tue, Oct 13, 2015 at 6:51 AM -0700, "Thomas Wiecki" <notifications@github.com> wrote:












In pyfolio/plotting.py:
> @@ -641,7 +649,25 @@ def plot_rolling_returns(
>              ax.fill_between(cone_df_live.index,
>                              cone_df_live.sd_down,
>                              cone_df_live.sd_up,
> -                            color='red', alpha=0.30)
> +                            color='steelblue', alpha=0.25)
> +
> +            if type(cone_std) is list:



There might be a more elegant way to do this. I'll think abit on it.

pyfolio/plotting.py
twiecki: Instead of having separate logic for list and float case. I would convert every float to a list, then have one loop.

pyfolio/plotting.py
twiecki: Why can't we generate the cones in one loop?

pyfolio/plotting.py
```diff
@@ -641,7 +649,25 @@ def plot_rolling_returns(
...
+                for cone_i in list(range(1, len(cone_std))):
```
twiecki: `for cone_std_i in cone_std:`

pyfolio/plotting.py
```diff
@@ -641,7 +649,25 @@ def plot_rolling_returns(
...
+                    cone_df = timeseries.cone_rolling(
+                        returns,
+                        num_stdev=cone_std[cone_i],
```
twiecki: `num_stdev=cone_std_i,`

pyfolio/plotting.py
```diff
@@ -641,7 +649,25 @@ def plot_rolling_returns(
...
+                    cone_df_fit = cone_df[cone_df.index < live_start_date]
+
+                    cone_df_live = cone_df[cone_df.index > live_start_date]
+                    temp_keep_days = cone_df_live.index < returns.index[-1]
+                    cone_df_live = cone_df_live[temp_keep_days]
+
+                    ax.fill_between(cone_df_live.index,
+                                    cone_df_live.sd_down,
+                                    cone_df_live.sd_up,
+                                    color='steelblue', alpha=0.25)
```
twiecki: I would take this whole logic and put it into a local function (`def draw_cone()`) that gets called inside the loop.

justinlent: I basically did the quickest implementation to touch the least amount of code so yea it can definitely be cleaned up further. I figured it's not that computationally expensive so went to simple rather than elegant route

**Sent from iPhone**




On Tue, Oct 13, 2015 at 1:45 PM -0700, "Thomas Wiecki" <notifications@github.com> wrote:












In pyfolio/plotting.py:
> +                for cone_i in list(range(1, len(cone_std))):
> +                    cone_df = timeseries.cone_rolling(
> +                        returns,
> +                        num_stdev=cone_std[cone_i],
> +                        cone_fit_end_date=live_start_date)
> +
> +                    cone_df_fit = cone_df[cone_df.index < live_start_date]
> +
> +                    cone_df_live = cone_df[cone_df.index > live_start_date]
> +                    temp_keep_days = cone_df_live.index < returns.index[-1]
> +                    cone_df_live = cone_df_live[temp_keep_days]
> +
> +                    ax.fill_between(cone_df_live.index,
> +                                    cone_df_live.sd_down,
> +                                    cone_df_live.sd_up,
> +                                    color='steelblue', alpha=0.25)



I would take this whole logic and put it into a local function (def draw_cone()) that gets called inside the loop.

twiecki: I think it's worth our time to go the elegant route. Technical debt has a very high interest rate.

justinlent: Ok that makes sense -  Yeah I can do this tonight most likely.

justinlent: I committed the update that factors out the cone rendering into a local draw_cone. definitely cleaner! by cleaning it up it also let me easily turn off the cone in the volatility match case -- was only a 3 "word" change so I snuck it in here instead of its own PR--yeah I know not great practice ;)

**Sent from iPhone**




On Tue, Oct 13, 2015 at 1:58 PM -0700, "Thomas Wiecki" <notifications@github.com> wrote:












In pyfolio/plotting.py:
> +                for cone_i in list(range(1, len(cone_std))):
> +                    cone_df = timeseries.cone_rolling(
> +                        returns,
> +                        num_stdev=cone_std[cone_i],
> +                        cone_fit_end_date=live_start_date)
> +
> +                    cone_df_fit = cone_df[cone_df.index < live_start_date]
> +
> +                    cone_df_live = cone_df[cone_df.index > live_start_date]
> +                    temp_keep_days = cone_df_live.index < returns.index[-1]
> +                    cone_df_live = cone_df_live[temp_keep_days]
> +
> +                    ax.fill_between(cone_df_live.index,
> +                                    cone_df_live.sd_down,
> +                                    cone_df_live.sd_up,
> +                                    color='steelblue', alpha=0.25)



I think it's worth our time to go the elegant route. Technical debt has a very high interest rate.

justinlent:
I committed the update that factors out the cone rendering into a local draw_cone. definitely cleaner! by cleaning it up it also let me easily turn off the cone in the volatility match case -- was only a 3 "word" change so I snuck it in here instead of its own PR--yeah I know not great practice ;)

pyfolio/plotting.py
```diff
@@ -610,16 +629,19 @@ def plot_rolling_returns(
...
             label='Live', ax=ax, **kwargs)
 
-        if cone_std is not None:
-            cone_df = timeseries.cone_rolling(
-                returns,
-                num_stdev=cone_std,
-                cone_fit_end_date=live_start_date)
-
-            cone_df_fit = cone_df[cone_df.index < live_start_date]
+        if cone_std is not None and not volatility_match:
```
twiecki: I think we shouldn't prohibit a user from drawing a cone if they set volatility_match to true. Instead, we should pass `cone_std=None` in the `create_returns_tearsheet()` function.

pyfolio/plotting.py
```diff
@@ -610,16 +629,19 @@ def plot_rolling_returns(
...
+            # check to see if cone_std was passed as a single value and,
+            # if so, just convert to list automatically
+            if isinstance(cone_std, float):
+                cone_std_list = [cone_std]
+            else:
```
twiecki: I think you can just replace `cone_std` with the list and remove the `else` statement that way.

pyfolio/plotting.py
```diff
@@ -610,16 +629,19 @@ def plot_rolling_returns(
...
+                cone_std_list = cone_std
 
-            cone_df_live = cone_df[cone_df.index > live_start_date]
-            cone_df_live = cone_df_live[cone_df_live.index < returns.index[-1]]
+            for cone_i in cone_std_list:
```
twiecki: looks much better.

pyfolio/plotting.py
```diff
@@ -573,6 +574,24 @@ def plot_rolling_returns(
         The axes that were plotted on.
 
 """
+    def draw_cone(returns, num_stdev, live_start_date, ax):
```
twiecki: Any reason to prepend the variable with `num_` here?

pyfolio/plotting.py
```diff
@@ -573,6 +574,24 @@ def plot_rolling_returns(
...
+        cone_df = timeseries.cone_rolling(
+            returns,
+            num_stdev=num_stdev,
+            cone_fit_end_date=live_start_date)
+
+        cone_df_fit = cone_df[cone_df.index < live_start_date]
+
+        cone_df_live = cone_df[cone_df.index > live_start_date]
+        temp_keep_days = cone_df_live.index < returns.index[-1]
```
twiecki: I find this double-slicing with a temp variable a bit confusing. Can we combine that to a single slice with an `&`?

pyfolio/plotting.py
```diff
@@ -573,6 +574,24 @@ def plot_rolling_returns(
         The axes that were plotted on.
 
 """
+    def draw_cone(returns, num_stdev, live_start_date, ax):
+        cone_df = timeseries.cone_rolling(
+            returns,
+            num_stdev=num_stdev,
+            cone_fit_end_date=live_start_date)
+
+        cone_df_fit = cone_df[cone_df.index < live_start_date]
```
twiecki: maybe `cone_in_sample` and `cone_out_of_sample`?

pyfolio/plotting.py
justinlent: Because I wanted to be consistent with the parameter name in timeseries.cone_rolling() which is the only place it gets passed. Is it worth changing for simplicity, or worth keeping because better to be consistent with where it is used downstream? thoughts?

pyfolio/plotting.py
justinlent: I agree, that I hate the temp variable, but I couldn't figure out how to be PEP8 compliant without breaking it up. I kept getting the SublimeText PEP8 flaker complaining at me.  Maybe i need to lookup how to breakup lines more effectlvely with PEP8

pyfolio/plotting.py
justinlent: Thanks for pointing that out (I actually didn't realize i could re-use a parameter variable like that.  Is this only possible because the parameter does not have risk of actually just being a pointer to an upstream variable? I'm always confused as to when Python treats an input parameter as merely a pointer to the upstream variable)

pyfolio/plotting.py
justinlent: yep,+1

pyfolio/plotting.py
```diff
@@ -582,6 +600,11 @@ def plot_rolling_returns(
                          'factor_returns.')
     elif volatility_match and factor_returns is not None:
         bmark_vol = factor_returns.loc[returns.index].std()
+        # TO-DO: @tweicki 'returns' probably needs to get updated to:
```
twiecki: I agree, if volatility_match is True, returns need to be changed/re-assigned as you proposed. That will fix the vol match plot.

pyfolio/plotting.py
```diff
@@ -582,6 +600,11 @@ def plot_rolling_returns(
...
+        # (returns / returns.std()) * bmark_vol if we want to plot
+        # the cone on this later on.
+        # Will we need a temp variable to do this? Or can we just re-assign as
+        # returns = (returns / returns.std()) * bmark_vol
         df_cum_rets = timeseries.cum_returns(
```
twiecki: Can then get rid of this line.

pyfolio/tears.py
```diff
@@ -42,7 +42,7 @@ def create_full_tear_sheet(returns, positions=None, transactions=None,
                            benchmark_rets=None,
                            gross_lev=None,
                            live_start_date=None, bayesian=False,
-                           cone_std=1.0, set_context=True):
+                           cone_std=[1.0, 1.5, 2.0], set_context=True):
```
twiecki: Minor stylistic but I'd change these to be tuples.

justinlent:
@twiecki Do you think this is good to merge soon?  Looks like merge conflicts are cropping up unfortunately. I'm out this upcoming Friday and Monday so I can try address anything else you see necessary

twiecki:
Merged with https://github.com/quantopian/pyfolio/commit/0040047f5c847c55d376a5537fe3ff1a2bc84b00. Sorry this took a while.

PR #174, opened 2015-10-17
Add stochastic volatility to Bayesian tear sheet

sbussmann:
Adds the following three plots to Bayesian tear sheet:

1. Posterior PDF for log(sigma)
2. Posterior PDF for log(nu)
3. Stochastic volatility overlaid on returns.


pyfolio/tears.py
```diff
@@ -629,6 +629,27 @@ def create_bayesian_tear_sheet(returns, benchmark_rets=None,
...
     ax_beta.set_ylabel('Belief')
 
+    # run stochastic volatility model
+    trace_stoch_vol = bayesian.model_stoch_vol(df_train)
+
+    # plot log(sigma) and log(nu)
+    row += 1
+    ax_sigma_log = plt.subplot(gs[row, 0])
+    ax_nu_log = plt.subplot(gs[row, 1])
+    sigma_log = trace_stoch_vol['sigma_log']
```
twiecki: Maybe the below to be a separate function?

twiecki: On second thought, might not be worth it.

twiecki:
Looks great, thanks. Can you remove `pyfolio/examples/.ipynb_checkpoints/bayesian-checkpoint.ipynb` and reset `spy.csv`? Also, adding some text to the Bayesian NB that describes the additional outputs would be a great addition.

sbussmann:
If you are happy with these changes, I'm ready to close this pull request.  I'm going to start working on understanding the runtime of the stochastic volatility model as a function of the size of the returns dataset.

twiecki:
Looks great! One last thing: I think we should add the truncation to the last 400 days and display a warning that we're truncating if the data passed in is more than 400 days.

sbussmann:
I just implemented this change on my local working copy.  However, I only noticed a minor performance improvement.  Using 400 days of returns (instead of the full set of 858 days), running the stochastic volatility model requires 88 seconds instead of ~110 seconds.  This is on a Macbook Air from late 2013.  How much testing have you done as far as number of days of returns and time required to run the stochastic volatility model?

twiecki:
Not much, but I have seen it fail for a larger number. If it works just as well on 858 days maybe this is fine. It's not rare though to pass in a 10 year backtest which would be ~2520 days and I would expect the model to fail in that case. Worth a try though.

sbussmann:
I'm going to open a new branch to explore time required to run model_stoch_vol as a function of number of returns and to work on caching.

sbussmann:
Do you mean something other than "cutoff"?

twiecki:
No, something other than replacing `df_returns`.

twiecki:
Thanks! Can you rebase this off of master and we can merge.

sbussmann:
I don't have experience with rebasing in git.  From looking at the git book, I'm thinking I need to do something like git rebase pyfolio/master.  But I'm not sure, and it seems that this is an area that can cause trouble if done poorly.  So I'd rather make sure I get it right the first time.  Let me know if this sounds right to you.

twiecki:
Rebased and merged via https://github.com/quantopian/pyfolio/commit/15663a69cf48c9d0222af81cd9cb78831815e4ae. Thanks!

Issue #156, opened 2015-09-29
Wrap all internal calls to helpers functions in create_full_tearsheet with try/except


justinlent:
This would help ensure the tearsheet can generate all possible plots even in the event of a single plot failing or throwing an exception

PR #189, opened 2015-11-02
Wrap all internal calls to helpers functions in create_full_tear_sheet with try/except

MridulS:
https://github.com/quantopian/pyfolio/issues/156

@justinlent Should I print out about the plots that failed. I used `pass` for the moment.

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -142,34 +142,48 @@ def create_full_tear_sheet(returns, positions=None, transactions=None,
     else:
         unadjusted_returns = None
 
-    create_returns_tear_sheet(
-        returns,
-        live_start_date=live_start_date,
-        cone_std=cone_std,
-        benchmark_rets=benchmark_rets,
-        set_context=set_context)
+    try:
+        create_returns_tear_sheet(
+            returns,
+            live_start_date=live_start_date,
+            cone_std=cone_std,
+            benchmark_rets=benchmark_rets,
+            set_context=set_context)
+    except:
+        pass
 
-    create_interesting_times_tear_sheet(returns,
-                                        benchmark_rets=benchmark_rets,
-                                        set_context=set_context)
+    try:
+        create_interesting_times_tear_sheet(returns,
+                                            benchmark_rets=benchmark_rets,
+                                            set_context=set_context)
+    except:
+        pass
 
     if positions is not None:
-        create_position_tear_sheet(returns, positions,
-                                   gross_lev=gross_lev,
-                                   hide_positions=hide_positions,
-                                   set_context=set_context,
-                                   sector_mappings=sector_mappings)
-
+        try:
+            create_position_tear_sheet(returns, positions,
+                                       gross_lev=gross_lev,
+                                       hide_positions=hide_positions,
+                                       set_context=set_context,
+                                       sector_mappings=sector_mappings)
+        except:
+            pass
         if transactions is not None:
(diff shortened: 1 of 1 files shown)


justinlent:
Hi @MridulS ,  Thanks for contributing to pyfolio! I think 'pass' for the moment to very sensible, as I wouldn't want the tearsheet output to get cluttered with warning/info messages since it is very common practice to export a tearsheet to PDF for presentation purposes.

justinlent:
Hi @MridulS ,
I was just reviewing the code diff for this PR, and was wondering if you could expand your try/except logic to the actual body of tears.create_returns_tearsheet() as well?  Basically wrapping each "plotting" call, e.g.:  plotting.plot_*  with a try/except.

Each of those plots is important, so I would like to keep any that plot succuessfully, and not simply catch any exception thrown by the entire call to create_returns_tearsheet().  Though, definitely keep your existing work as it is as it is certainly helpful as well at the topmost-level. 

Thoughts?

twiecki:
I'm not sure about this. Certainly swallowing exceptions without any information is wrong as it will mask bugs and not give any indication as what to do to fix things. Given input data in the correct format, why should a certain plot break but not another? I'd rather make sure all plots work with the data you can pass in or have specific exceptions for special cases.

justinlent:
@twiecki I see your point, but I guess I'm torn.  I think we want pyfolio to be particularly user-friendly to the coding 'naive' user (e.g. the Finance guy who lives in Excel), so failing gracefully with no exceptions via passing over stuff that doesn't work seems right to me.  Obviously if we could catch every potential bad input and handle it properly that would be great, however given how restrictive pandas/matplotlib/etc is with handling timezones/DateTimeIndex's, etc, I think this will be difficult and frustrating for anyone not up the Python learning curve.  TBH even I had much difficulty just trying to pass in some naive daily returns with simple datestamps, until I spent some time figuring out how to modify timezone settings and normalize/localize DateTimeIndex's.  

Maybe we need a 'debug' setting/flag for developers/programming quants that operates in the better state which you've described (throwing exceptions, etc), but if it's set to 'false' (for the non-coder user) it just passes over exceptions and fails gracefully, but still generates all possible plots.

Thoughts?

twiecki:
I see your point and I agree. The timezone thing is a nuisance and it really doesn't matter for daily returns. However, if the input data is correct, all plots should work that rely on the same data. So I'd rather fix the data input problem rather than having individual plots break for unknown reasons.

For example, we could just check the input data for timezone info and adjust if it's not set and display a warning.

justinlent:
@twiecki Cool. Ok yeah I agree with your last reply.  So how do you want to handle this PR? Merge just the existing, or just close in favor of a new issue to address various data input problems?

MridulS:
It would be great if anyone could point out an example(data input) where the code breaks for a certain plot.
@justinlent I think we can close this PR, and work on data input problems.

twiecki:
@MridulS I agree, thanks again for putting this up, we appreciate your help.

What I'm imagining is a function like `datetime_index_to_utc(input)` which checks if the data is daily (in which case it should always be safe to convert) and whether it's already utc (in which case we don't have to do anything) and convert if it's not and display a warning saying "Input data not UTC, automatically converting. To avoid this warning, use `.tz_localize('utc') or .tz_convert('utc')` before passing it to pyfolio.

The next issue is probably that we will want to do that for all functions, so maybe we should create a decorator `sanitize_tz_inputs` that converts automatically.

Issue #191, opened 2015-11-03
Remove hard-coded "S&P500"


jkrauss82:
if providing individual benchmark_rets (e. g. ^GDAXI) some plots still show S&P500 in headline

Affected functions in plotting.py:
plot_rolling_beta
plot_rolling_returns

twiecki:
Thanks @jkrauss82. Want to submit a PR that fixed it?

twiecki:
Fixed by #192.

PR #192, opened 2015-11-03
removed hard-coded s&p500

jkrauss82:
instead using name attribute of pandas series now, concerns ticket #191


twiecki:
Thanks @jkrauss82. LGTM.

PR #195, opened 2015-11-03
ENH implement plot_monthly_returns_timeseries in plotting.py to rende…

justinlent:
…r monthly bars representing monthly returns of an algo

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -1344,3 +1344,48 @@ def show_worst_drawdown_periods(returns, top=5):
     drawdown_df['net drawdown in %'] = list(
         map(utils.round_two_dec_places, drawdown_df['net drawdown in %']))
     print(drawdown_df.sort('net drawdown in %', ascending=False))
+
+
+def plot_monthly_returns_timeseries(returns, ax=None, **kwargs):
+    """
+    Plots monthly returns as a timeseries.
+
+    Parameters
+    ----------
+    returns : pd.Series
+        Daily returns of the strategy, noncumulative.
+         - See full explanation in tears.create_full_tear_sheet.
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    **kwargs, optional
+        Passed to seaborn plotting function.
+
+    Returns
+    -------
+    ax : matplotlib.Axes
+        The axes that were plotted on.
+    """
+
+    if ax is None:
+        ax = plt.gca()
+
+    monthly_ret_table = timeseries.aggregate_returns(returns, 'monthly')
+    monthly_ret_table = monthly_ret_table.reset_index()
+    monthly_ret_table.columns = ['year', 'month', 'returns']
+
+    # Generate month-year labels for the x-axis corresponding to
+    # the returns plotted
+    date_labels = [str(i[1]) + ' - ' + str(i[0]) for i in
+                   zip(monthly_ret_table.year.values,
+                   monthly_ret_table.month.values)]
+
+    sns.barplot(x=date_labels,
+                y=monthly_ret_table.returns,
+                color='steelblue')
+    locs, labels = plt.xticks()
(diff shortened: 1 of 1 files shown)


justinlent:
I want to add this plot as a default in tears.create_returns_tearsheet().

justinlent:
So, right now on backtests with many years of returns, the x-axis on this monthly return's plot is pretty cluttered when plotting every month's label. I will revise to only plot a label for each January

justinlent:
Here's an example of what this feature looks like when rendered on an algo's returns:

image

justinlent:
Here's an example of what this feature looks like when rendered on an algo's returns:

image

justinlent:
@twiecki let me know if you think we can merge this.  It only involves additions to the code base, and no modifications to anything existing, so pretty low risk. Let me know if it looks good to you.

pyfolio/plotting.py
```diff
@@ -1344,3 +1344,63 @@ def show_worst_drawdown_periods(returns, top=5):
...
+    if ax is None:
+        ax = plt.gca()
+
+    monthly_ret_table = timeseries.aggregate_returns(returns, 'monthly')
+    monthly_ret_table = monthly_ret_table.reset_index()
+    monthly_ret_table.columns = ['year', 'month', 'returns']
+
+    # Generate month-year labels for the x-axis corresponding to
+    # the returns plotted
+    date_labels = [str(i[1]) + ' - ' + str(i[0]) for i in
```
twiecki: What's wrong with the default labels of this plot? I suppose there's too many? How about just thinning those. The current approach feels a bit brittle.

justinlent: yep that's right, the defaults are super ugly because it prints the month/day/year of each rendered monthly bar and the plot becomes illegible if more than ~30 bars. 
As well, the defaults include the last business _day_ of the month in each label (which just adds to the cluttered visual because the last business day of each month jumps around, 29, 30, 31, etc). My opinion is that including the _day_ in the label is not really proper when labeling a monthly value, as well.
Pandas does not have a monthly frequency unit so I could not up-sample the DateTimeIndex to make it more elegant (which would have been best).
Given the above, and my desire to have an easily interpreted plot, how would you suggest improved pruning of the labels? I actually adapted my implementation from an accepted StackOverflow response
**Sent from iPhone**

    _____________________________

justinlent: @twiecki ...Actually, my bad, looks like Pandas _does_ support a Monthly unit, under the concept of a Period. so timestamps can be converted to _periods_ to drop the concept of a "day" in monthly data and have plots generated properly. I'll go ahead and re-implement...  Ohhh Pandas, so convoluted sometimes...

**Sent from iPhone**




On Wed, Nov 4, 2015 at 4:10 AM -0800, "Thomas Wiecki" <notifications@github.com> wrote:












In pyfolio/plotting.py:
> +    Returns
> +    -------
> +    ax : matplotlib.Axes
> +        The axes that were plotted on.
> +    """
> +
> +    if ax is None:
> +        ax = plt.gca()
> +
> +    monthly_ret_table = timeseries.aggregate_returns(returns, 'monthly')
> +    monthly_ret_table = monthly_ret_table.reset_index()
> +    monthly_ret_table.columns = ['year', 'month', 'returns']
> +
> +    # Generate month-year labels for the x-axis corresponding to
> +    # the returns plotted
> +    date_labels = [str(i[1]) + ' - ' + str(i[0]) for i in



What's wrong with the default labels of this plot? I suppose there's too many? How about just thinning those. The current approach feels a bit brittle.

justinlent:
@twiecki I re-implemented using PeriodIndex - much cleaner implementation for sure.  As well rendering some yearly boundary dashed-lines for viewing convenience.  
image

twiecki:
@justinlent Looks great!

PR #196, opened 2015-11-05
information ration optionally takes benchmark

cmorgan:
Returns series may already be benchmark adjusted.
Has side effect of keeping perf_stats running.


twiecki:
Can you add that to the doc string?

cmorgan:
done

twiecki:
Hm, I start to wonder, why should we support that specific case?

cmorgan:
maybe it's better to impose this standardisation on the API, closing

twiecki:
OK, I agree. Thanks for the suggestion though, we do appreciate your contributions.

Issue #197, opened 2015-11-05
pos.get_percent_alloc not working correctly with short positions (?)


jkrauss82:
The function mentioned in the subject is making values absolute before calculating their relative weight in the portfolio. However this leads to incorrect weights for short positions when assuming that you actually increase your cash when short selling a stock.

As I take from the positions example in tears.py

index         'AAPL'         'MSFT'          cash
2004-01-09    13939.3800     -14012.9930     711.5585
2004-01-12    14492.6300     -14624.8700     27.1821
2004-01-13    -13853.2800    13653.6400      -43.6375

the rationale of pyfolio might be different (I assume you do not want to be leveraged by 21 times in the AAPL position for this example, actually get_percent_alloc returns 0.486304 for this position). However, in reality (e. g. on Interactive Brokers) it does not work that way - what is your view on that matter?

ACampbell22:
I think you are correct. We should be using the net liquidation value (`net_liquidation = longs - abs(shorts) + cash`) as the denominator in get_percent_alloc. Currently we are double counting shorts. 

We should also probably change that positions example to depict a more likely scenario where cash reflects the presence of short positions under the assumption of < 3x leverage.

Thanks for the heads up!

humdings:
@jkrauss82 Take a look at `pos.get_long_short_pos`, that should calculate the long and short allocations how you are expecting.

jkrauss82:
Well actually I got it working simply by removing the `.abs()` in line 37 of pos.py. The way we create our input data then works fine for the rest of the process and percentage values come out correct.

I am not familiar with all of the pyfolio code so I do not know if other functions depend on `pos.get_percent_alloc` and depend on the old way?

twiecki:
Closed with https://github.com/quantopian/pyfolio/pull/201. Thanks for reporting @jkrauss82.

PR #201, opened 2015-11-11
BUG use net liquidation value as the denominator in percent allocation calculation

ACampbell22:
Re: https://github.com/quantopian/pyfolio/issues/197


ACampbell22:
@humdings does this look ok?

twiecki:
Waiting for @humdings feedback. Other fix looks good.

humdings:
@ACampbell22 Looks good to me, this will account for the leverage now.

Issue #203, opened 2015-11-12
Add cents/share profit metric


twiecki:
How much profit/loss was made per share.

PR #205, opened 2015-11-13
WIP Add cents-per-share calculation.

twiecki:
WIP -- DO NOT MERGE

Implements https://github.com/quantopian/pyfolio/issues/203

The issue is that it requires PnL which we do not use anywhere. We could compute it from returns and capital base, but we don't use capital base anywhere either. So here I'm then computing it by summing up the daily positions. Not sure that's the best way to go about doing this.

Ideas?

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -1364,7 +1364,7 @@ def information_ratio(returns, benchmark_returns):
     return np.mean(active_return) / tracking_error
 
 
-def cents_per_share(returns, capital_base, txn_volume):
+def cents_per_share(returns, positions, txn_volume):
     """
     Calculates average cents per share earned (or lost).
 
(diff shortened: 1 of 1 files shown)


justinlent:
Not sure what the original discussion was surrounding this metric, but I think this should be based purely on the buy/sell _transactions_.  So we have to choose either FIFO or LIFO  (I would probably choose FIFO) and than match roundturn transactions based on it.  and the cents-per-share calculation is just the net difference between the buy/sell FIFO mapped roundturn transactions.

Do you think that makes sense?

twiecki:
Do you mean to tack this on to the round-trip analysis then? It seems like we could calculate this more easily by just averaging the PnL / #shares traded.

justinlent:
Yea I think your approach makes sense of PnL/#shares traded.  Tacking onto roundtrip analysis seems like a great place for this. If it's easy to describe, how is what you've implemented in this branch different from just PnL/Shares traded?

twiecki:
That's what it's supposed to compute.

twiecki:
Closing for the time being. Will reopen when I have tests.

PR #210, opened 2015-11-16
ENH Added round trip tearheet and supporting functions

a-campbell:
- Adds ability to split trades
- Adds ability to append "closing" trades to complete round trips open at the end of the backtest
- Adds various plotting and printing functions to visualize round trip duration, frequency, and profitability
- Adds ability to display profit attribution by sector.
- Tests for round trip extraction functions.

screenshot 2015-11-16 18 46 25


ACampbell22:
@twiecki should you review this since you wrote a lot of this code?

ssanderson:
> We also need more unittests I think.
:+1: 

Also noticed while browsing: many of the new functions' docstrings report that they accept `**kwargs` and forward them to seaborn, which appears not to be true.

pyfolio/plotting.py
```diff
@@ -1402,3 +1404,116 @@ def cumulate_returns(x):
...
+    ax.set_yticklabels(symbols)
+
+    red_line = mlines.Line2D([], [], color='r', label='Long')
+    blue_line = mlines.Line2D([], [], color='b', label='Short')
+    ax.legend(handles=[red_line, blue_line], loc=0)
+
+    return ax
+
+
+def show_profit_attribtion(round_trips):
```
twiecki: attribution

pyfolio/plotting.py
```diff
@@ -1402,3 +1404,116 @@ def cumulate_returns(x):
...
+    Plots timespans and directions of round trip trades.
+
+    Parameters
+    ----------
+    round_trips : pd.DataFrame
+        DataFrame with one row per round trip trade.
+        - See full explanation in txn.extract_round_trips
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    **kwargs, optional
```
twiecki: Doesn't take kwargs.

twiecki:
`autopep` is your friend :).

Also, I think an example NB with some description would go a long way. Really like where this is going.

llllllllll:
You might want to consider gzipping the test data. If there is a lot it will make cloning much slower, especially when you don't have great internet.

ACampbell22:
good call

pyfolio/round_trips.py
```diff
@@ -0,0 +1,248 @@
...
+    return round_trips
+
+
+def split_trades(transactions):
+    """
+    Splits transactions that cause total position amount to cross zero.
+    In other words, separates of the closing of one short/long position
+    with the opening of a new long/short position.
+
+    For example, the second transaction in this transactions DataFrame
```
twiecki: Very nice/clear description.

twiecki:
LGTM. I think there are still the csv files present here.

Issue #213, opened 2015-11-19
Fix Sharpe ratio calculation


twiecki:
Should be: mean(daily returns - risk_free) / std(daily returns - risk_free) * np.sqrt(252)

Adding for reference:
https://web.stanford.edu/~wfsharpe/art/sr/sr.htm

CC @justinlent

justinlent:
+1
Outstanding question is how to handle the risk_free rate timeseries when it is commonly pulled in as an annual yield.

E.g.  If the daily value in the risk_free timeseries is 4% annual yield, is the daily yield:
a)  4% / 252
b) (1 + 4%)^(1/252) - 1

My thought is (b)

twiecki:
Yep, agree with b:
```python

PR #219, opened 2015-11-24
BUG Fix Sharpe ratio calculation.

twiecki:
Fixes https://github.com/quantopian/pyfolio/issues/213


twiecki:
This does not yet include the risk-free but allows for adding it later.

justinlent:
+1
it looks GTG

Issue #222, opened 2015-11-29
Rolling Beta plot fails if benchmark returns series name isn't a string.


ssanderson:
For example, on Q Research, the name of the benchmark returns series is a Zipline Equity object, resulting in this traceback:
```
/usr/local/lib/python2.7/dist-packages/pyfolio/tears.pyc in create_full_tear_sheet(returns, positions, transactions, benchmark_rets, gross_lev, slippage, live_start_date, bayesian, hide_positions, sector_mappings, cone_std, set_context)
    148         cone_std=cone_std,
    149         benchmark_rets=benchmark_rets,
--> 150         set_context=set_context)
    151 
    152     create_interesting_times_tear_sheet(returns,

/usr/local/lib/python2.7/dist-packages/pyfolio/plotting.pyc in call_w_context(*args, **kwargs)
     42         if set_context:
     43             with context():
---> 44                 return func(*args, **kwargs)
     45         else:
     46             return func(*args, **kwargs)

/usr/local/lib/python2.7/dist-packages/pyfolio/tears.pyc in create_returns_tear_sheet(returns, live_start_date, cone_std, benchmark_rets, return_fig)
    273 
    274     plotting.plot_rolling_beta(
--> 275         returns, benchmark_rets, ax=ax_rolling_beta)
    276 
    277     plotting.plot_rolling_sharpe(

/usr/local/lib/python2.7/dist-packages/pyfolio/plotting.pyc in plot_rolling_beta(returns, factor_returns, legend_loc, ax, **kwargs)
    701     ax.yaxis.set_major_formatter(FuncFormatter(y_axis_formatter))
    702 
--> 703     ax.set_title("Rolling Portfolio Beta to " + factor_returns.name)
    704     ax.set_ylabel('Beta')
    705     rb_1 = timeseries.rolling_beta(

TypeError: cannot concatenate 'str' and 'zipline.assets._assets.Equity' objects
```

In general, I don't think pyfolio should be expecting that the returns timeseries has a string for a name.

PR #223, opened 2015-11-29
Name might not be a string

ssanderson:
Fixes https://github.com/quantopian/pyfolio/issues/222


ssanderson:
@twiecki this is a counterproposal to your downstream fix here: https://github.com/quantopian/qexec/pull/8520.

twiecki:
LGTM. Thanks. Now only have to update research.

PR #228, opened 2015-12-01
MAINT Pass factor_returns to perf_stats and compute alpha, beta and information ratio.

twiecki:


twiecki:
Also fixes another breaking bug with calling sharpe_ratio() with the old arguments.

Issue #178, opened 2015-10-21
timeseries.cone_rolling needs refactoring


twiecki:
@justinlent and I just looked at the code in `cone_rolling`. While functional, as we're relying on that code more and more it should be more robust. There seems to be an issue with <= logics, the warmup-period is obsolete, it's not tested etc. Probably a clean rewrite is the best path forward.

twiecki:
@ACampbell22 Is that something you could help with? I think the complete code should be rewritten from first principles.

twiecki:
Moreover, a nice addition would be if it were possible to output the cone in a dataframe so that we can add automatic checks if a cone was breached.

a-campbell:
Started looking at this last night. Will get going on a refactor soon.

twiecki:
:+1:

a-campbell:
Do we want to continue to support the extend_fit_trend=False (refit linear cone with ever additional day of data) option? It adds a lot of complexity and isn't used in the full tear sheet.

twiecki:
Agreed, lets drop that.

a-campbell:
Is it right to use the slope of the line fit to the backtest cumulative returns as the slope of the midline of the future cone projection? If an algo returns 10 bps per day, its equity curve will have a much higher linear fit slope over a 13 year backtest than it will over a 1 year backtest (compounding). 

Would it be more correct to project the backtest CAGR forward? 

I am trying to make the "cone bounds" function as generalizable as possible. Should be applicable to any OOS starting portfolio value.

justinlent:
Hey Andrew, yes I agree simply applying the CAGR from the backtest as the OOS trend projection is a better, generalized implementation. I actually used that is a way to check the math of the regression approach if I recall since they should be equivalent when the regression is done with the intercept term (which I think is how it was implemented originally). 


**Sent from iPhone**

    _____________________________

justinlent:
@a-campbell ,
After sleeping on it, I think we'll want to try out both applications (the mean daily return, and the CAGR approach) to see what feels the most correct.  I'm a bit torn now that I think about it. Since the cone is calculated _everyday_ for application of risk monitoring in more or less _real-time_ I feel like we should be using the simple mean returns.  As I see it we have 2 approaches  up for debate for drawing the slop of the line that bisects the cone which are dependent on choosing how to compute the _OOS daily return_ projection. So should the daily return be computed as:

a)  mean(backtest_daily_returns)
[...]

PR #233, opened 2015-12-02
ENH Replace linear cone with bootstrapped non-parametric cone

a-campbell:
The old linear cone had several troubling assumptions (normal returns, use of arithmetic-like return averaging in the backtest, linearity of future returns, starting capital of oos == fixed starting value of the cone).

This new bootstrapped cone makes fewer assumptions by building its forward projection from samples drawn directly from the in-sample daily returns.  This direct sampling makes the bootstrapped cone non-parametric, it does not assume backtest returns fit into any particular distribution.

Discussion here: https://github.com/quantopian/pyfolio/issues/178

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -521,21 +521,21 @@ def show_perf_stats(returns, factor_returns, live_start_date=None):
     print(perf_stats)
 
 
-def plot_rolling_returns(
-        returns,
-        factor_returns=None,
-        live_start_date=None,
-        cone_std=None,
-        legend_loc='best',
-        volatility_match=False,
-        ax=None, **kwargs):
-    """Plots cumulative rolling returns versus some benchmarks'.
+def plot_rolling_returns(returns,
+                         factor_returns=None,
+                         live_start_date=None,
+                         cone_std=None,
+                         legend_loc='best',
+                         volatility_match=False,
+                         ax=None, **kwargs):
+    """
+    Plots cumulative rolling returns versus some benchmarks'.
 
     Backtest returns are in green, and out-of-sample (live trading)
     returns are in red.
 
-    Additionally, a linear cone plot may be added to the out-of-sample
-    returns region.
+    Additionally, a non-parametric cone plot may be added to the
+    out-of-sample returns region.
 
     Parameters
     ----------
@@ -546,13 +546,12 @@ def plot_rolling_returns(
         Daily noncumulative returns of a risk factor.
          - This is in the same style as returns.
     live_start_date : datetime, optional
-        The point in time when the strategy began live trading, after
-        its backtest period.
+        The date when the strategy began live trading, after
+        its backtest period. This date should be normalized.
     cone_std : float, or tuple, optional
         If float, The standard deviation to use for the cone plots.
         If tuple, Tuple of standard deviation values to use for the cone plots
-         - The cone is a normal distribution with this standard deviation
-             centered around a linear regression.
+         - See timeseries.forecast_cone_bounds for more details.
     legend_loc : matplotlib.loc, optional
         The location of the legend on the plot.
     volatility_match : bool, optional
@@ -570,26 +569,11 @@ def plot_rolling_returns(
         The axes that were plotted on.
 
 """
-    def draw_cone(returns, num_stdev, live_start_date, ax):
-        cone_df = timeseries.cone_rolling(
-            returns,
-            num_stdev=num_stdev,
-            cone_fit_end_date=live_start_date)
-
-        cone_in_sample = cone_df[cone_df.index < live_start_date]
-        cone_out_of_sample = cone_df[cone_df.index > live_start_date]
-        cone_out_of_sample = cone_out_of_sample[
-            cone_out_of_sample.index < returns.index[-1]]
-
-        ax.fill_between(cone_out_of_sample.index,
-                        cone_out_of_sample.sd_down,
-                        cone_out_of_sample.sd_up,
-                        color='steelblue', alpha=0.25)
-
-        return cone_in_sample, cone_out_of_sample
-
     if ax is None:
         ax = plt.gca()
+
+    ax.set_ylabel('Cumulative returns')
+    ax.set_xlabel('')
 
     if volatility_match and factor_returns is None:
         raise ValueError('volatility_match requires passing of'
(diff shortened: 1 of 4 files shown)


a-campbell:
@twiecki I think this might be good to go.

pyfolio/timeseries.py
```diff
@@ -985,129 +985,60 @@ def rolling_sharpe(returns, rolling_sharpe_window):
...
-        num_stdev=1.0,
-        warm_up_days_pct=0.5,
-        std_scale_factor=APPROX_BDAYS_PER_YEAR,
-        update_std_oos_rolling=False,
-        cone_fit_end_date=None,
-        extend_fit_trend=True,
-        create_future_cone=True):
-    """Computes a rolling cone to place in the cumulative returns
-    plot. See plotting.plot_rolling_returns.
+def forecast_cone_bounds(is_returns, num_days, cone_std,
```
twiecki: we might want to add more, so maybe `forecast_cone_bootstrap`?

twiecki: Also could provide default cone_std values of 1, 1.5, and 2.

pyfolio/timeseries.py
```diff
@@ -985,129 +985,60 @@ def rolling_sharpe(returns, rolling_sharpe_window):
...
-        future_cone = pd.DataFrame(index=list(map(np.datetime64,
-                                                  future_cone_dates)),
-                                   data={'perf': temp_line,
-                                         'line': temp_line,
-                                         'sd_up': temp_sd_up,
-                                         'sd_down': temp_sd_down})
-
-        perf_ts_r = perf_ts_r.append(future_cone)
-
-    return perf_ts_r
```
twiecki: So great to see all that complexity go.

pyfolio/plotting.py
```diff
@@ -598,65 +582,55 @@ def draw_cone(returns, num_stdev, live_start_date, ax):
...
-                **kwargs)
-            cone_out_of_sample['line'].plot(
-                ax=ax,
-                ls='--',
-                label='Predicted trend',
-                lw=2,
-                color='red',
-                alpha=0.7,
-                **kwargs)
+            is_daily_returns = is_cum_returns.pct_change().dropna()
```
twiecki: Any reason not to use the returns directly? E.g. `is_daily_returns = returns.loc[returns.index < live_start_date]`

Also, I think you can drop the `daily` in the name for consistency as we don't use it elsewhere.

pyfolio/plotting.py
```diff
@@ -598,65 +582,55 @@ def draw_cone(returns, num_stdev, live_start_date, ax):
...
+            cone_bounds = timeseries.forecast_cone_bounds(
+                is_daily_returns,
+                len(oos_cum_returns),
+                cone_std,
+                starting_value=is_cum_returns[-1])
```
twiecki: Perfect, this is fixing the bug we had before.

pyfolio/plotting.py
```diff
@@ -598,65 +582,55 @@ def draw_cone(returns, num_stdev, live_start_date, ax):
...
-            cone_out_of_sample['line'].plot(
-                ax=ax,
-                ls='--',
-                label='Predicted trend',
-                lw=2,
-                color='red',
-                alpha=0.7,
-                **kwargs)
+            is_daily_returns = is_cum_returns.pct_change().dropna()
+            cone_bounds = timeseries.forecast_cone_bounds(
```
twiecki: I wonder if we should add this function as a default parameter; allowing users to pass in their own cone-generating function.

pyfolio/plotting.py
```diff
@@ -598,65 +582,55 @@ def draw_cone(returns, num_stdev, live_start_date, ax):
...
+                is_daily_returns,
+                len(oos_cum_returns),
+                cone_std,
+                starting_value=is_cum_returns[-1])
+
+            cone_bounds = cone_bounds.set_index(oos_cum_returns.index)
+
+            for std in cone_std:
+                ax.fill_between(cone_bounds.index,
+                                cone_bounds[str(std)],
```
twiecki: Any reason to use string here?

pyfolio/timeseries.py
```diff
@@ -985,129 +985,60 @@ def rolling_sharpe(returns, rolling_sharpe_window):
...
+        Number of days to project the probability cone forward.
+    cone_std : int, float, or list of int/float
+        Number of standard devations to use in the boundaries of
+        the cone. If multiple values are passed, cone bounds will
+        be generated for each value.
+    starting_value : int or float
+        Starting value of the out of sample period.
+    num_samples : int
+        Number of num_days length samples to draw from the in-sample
+        daily returns.
```
twiecki: This reads a bit clunky. Maybe also add that increasing this number will increase accuracy of the cone.

pyfolio/timeseries.py
```diff
@@ -985,129 +985,60 @@ def rolling_sharpe(returns, rolling_sharpe_window):
...
-    perf_ts_r.columns = ['perf']
+    Returns
+    -------
+    pd.DataFrame
+        Contains upper and lower cone boundaries. Column names are
+        strings corresponding to the number of standard devations
+        above (positive) or below (negative) the projected mean
+        cumulative returns.
+    """
+    np.random.seed(100)
```
twiecki: I don't think we should set the seed by default. Instead, we could make it an argument that defaults to None.

pyfolio/timeseries.py
```diff
@@ -985,129 +985,60 @@ def rolling_sharpe(returns, rolling_sharpe_window):
...
-    new_cone_day_scale_factor = int(1)
-    oos_intercept_shift = perf_ts_r.perf[-1] - perf_ts_r.line[-1]
+    if isinstance(cone_std, (float, int)):
+        cone_std = [cone_std]
 
-    # make the cone for the out-of-sample/live papertrading period
-    for i in cone_end_rets.index:
-        returns = input_rets[:i]
-        perf_ts = cum_returns(returns, 1)
+    cone_bounds = pd.DataFrame()
```
twiecki: Already initializing columns and index and then using `.loc` below would make this a bit more readable.

twiecki:
Looks great overall but see minor comments.

pyfolio/timeseries.py
a-campbell: In the interest of producing deterministic tear sheets, might we want to set a default seed?

twiecki: That's probably fine.

pyfolio/plotting.py
a-campbell: Ah, yea we can explicitly define the columns as floats. When we don't explicitly say `columns=pd.Float64Index([])` and the first column name is float(1) or int(1), pandas assumes the columns are a Int64Index and blows up when we try to insert a column named float(1.5).

pyfolio/timeseries.py
a-campbell: changed

pyfolio/timeseries.py
a-campbell: changed

pyfolio/plotting.py
a-campbell: great idea. made the change.

pyfolio/plotting.py
a-campbell: good call. made the change.

pyfolio/timeseries.py
a-campbell: made the change

twiecki:
Merged with https://github.com/quantopian/pyfolio/commit/9cc98eb4c3535d8488bd8f10532704b66a1cf35c :+1:

Issue #218, opened 2015-11-24
add long/short position concentration plot to tearsheet


Jstauth:
![12_yr_summary_3](https://cloud.githubusercontent.com/assets/5178573/11378521/9daeb78c-92b9-11e5-9587-de0d245e4946.png)

Andrew built this awesome long/short position concentration plot that we've looked at previously. Should make it into pyfolio I think as it is generally applicable and useful across many different algos.

PR #237, opened 2015-12-03
ENH Plot max and median long/short exposures

a-campbell:
Issue: https://github.com/quantopian/pyfolio/issues/218

screenshot 2015-12-03 18 21 13

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -919,6 +919,37 @@ def show_and_plot_top_positions(returns, positions_alloc,
         return ax
 
 
+def plot_max_median_position_concentration(positions, ax=None, **kwargs):
+    """
+    Plots the max and median of long and short position concentrations
+    over the time.
+
+    Parameters
+    ----------
+    positions : pd.DataFrame
+        The positions that the strategy takes over time.
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+
+    Returns
+    -------
+    ax : matplotlib.Axes
+        The axes that were plotted on.
+    """
+    if ax is None:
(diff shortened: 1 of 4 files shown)


a-campbell:
A misleading part of this plot is the fact that median takes the average two position concentrations when there are an even number of longs or shorts.

pyfolio/plotting.py
```diff
@@ -936,6 +936,38 @@ def show_and_plot_top_positions(returns, positions_alloc,
...
+    Plots the max and median of long and short position concentrations
+    over the time.
+
+    Parameters
+    ----------
+    positions : pd.DataFrame
+        The positions that the strategy takes over time.
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    **kwargs, optional
```
twiecki: Left-over.

twiecki:
> A potentially misleading part of this plot is the fact that median takes the average two position concentrations when there are an even number of longs or shorts.

It should take the average between the two center values if there's an even number but that's fine and expected.

Otherwise LGTM.

pyfolio/pos.py
```diff
@@ -97,6 +97,37 @@ def get_top_long_short_abs(positions, top=10):
     return df_top_long, df_top_short, df_top_abs
 
 
+def get_max_median_position_conc(positions):
```
twiecki: `get_max_median_position_concentration`

a-campbell:
Should be good to go now.

pyfolio/plotting.py
twiecki: Still doesn't do anything with kwars below.

PR #242, opened 2015-12-08
BUG: fix caching logic

llllllllll:


pyfolio/utils.py
```diff
@@ -157,11 +157,23 @@ def get_returns_cached(filepath, update_func, latest_dt, **kwargs):
         returns = update_func(**kwargs)
         try:
             mkdir(cache_dir())
+        except OSError as e:
```
ssanderson: This probably wants to be a function.  We use this in the Q codebase downstream:

```
import os
from errno import EEXIST


def ensure_directory(path):
    """
    Ensure that a directory named "path" exists.
    """
    try:
        os.makedirs(path)
    except OSError as exc:
        if exc.errno == EEXIST and os.path.isdir(path):
            pass
        else:
            raise
```

llllllllll: ;_; python3 supports an `exists_ok` flag for this.

pyfolio/utils.py
```diff
@@ -156,12 +167,24 @@ def get_returns_cached(filepath, update_func, latest_dt, **kwargs):
...
+                UserWarning,
+            )
+
+        try:
             returns.to_csv(filepath)
-        except IOError as e:
-            warnings.warn('Could not update cache {}.'
-                          'Exception: {}'.format(filepath, e),
-                          UserWarning)
+        except Exception as e:
```
ssanderson: Why did this go from `IOError` to any exception?

llllllllll: I think that we should suppress any errors that come out of `to_csv`. This is non-fatal behavior and the warning is sufficient to signal that the caching has failed.

ssanderson: Do we want to be catching things like `MemoryError` though?  In `zipline.data.loader` we do this:
```
    # Stuff involving pandas to_csv/from_csv
    except (OSError, IOError, ValueError) as e:
        # These can all be raised by various versions of pandas on various
        # classes of malformed input.  Treat them all as cache misses.
        logger.info(
            "Loading data for {path} failed with error [{error}].".format(
                path=path, error=e,
            )
        )
```

llllllllll: I don't see why we wouldn't want to catch that. If the failure is in `to_csv` that does not affect the rest of the application. If the memory error persists and causes a required component to fail, then let that raise at the point of failure.

ssanderson: My gut is that we should try as little as possible to suppress unexpected errors, since the failure mode there tends to be harder-to-debug failures elsewhere.  If you feel strongly about this though, then you can merge as-is.

llllllllll:
merging on pass

PR #245, opened 2015-12-09
MAINT: remove unused data files

llllllllll:
These files make cloning considerably slower. They are also no longer looked up in the repo tree so they are unused. I could not find anything that referenced `example_positions.csv` either in code or the docs so I figure it must not be used.


twiecki:
LGTM, thanks @llllllllll.

PR #250, opened 2015-12-09
ENH: Suppress seaborn import warning

llllllllll:
currently we raise a warning on import. This is actually coming out of matplotlib because of something in the import of seaborn
```python
In [1]: import pyfolio
/home/yui/.virtualenvs/pyfolio3/lib/python3.4/site-packages/matplotlib/__init__.py:872: UserWarning: axes.color_cycle is deprecated and replaced with axes.prop_cycle; please use the latter.
  warnings.warn(self.msg_depr % (key, alt_key))
```

This suppresses that warning because it is not directly our fault.


twiecki:
:+1:

Issue #208, opened 2015-11-16
Add bootstrap to gain confidence in perf stats


twiecki:
All perf stats like Sharpe should be computed along with a bootstrap analysis to get a sense of uncertainty.

twiecki:
https://en.wikipedia.org/wiki/Bootstrapping_(statistics)

twiecki:
Closed by https://github.com/quantopian/pyfolio/pull/261

PR #261, opened 2015-12-18
ENH Bootstrap analysis

twiecki:
Addresses https://github.com/quantopian/pyfolio/issues/208. Depends on https://github.com/quantopian/pyfolio/pull/262.

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -811,7 +811,9 @@ def calc_bootstrap(func, returns, *args, **kwargs):
     factor_returns : pd.Series (optional)
         Daily noncumulative returns of the benchmark.
          - This is in the same style as returns.
-
+    n_samples : int (optional)
+        Number of bootstrap samples to draw. Default is 1000.
+        Increasing this will lead to more stable / accurate estimates.
     Returns
     -------
     numpy.ndarray
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -824,18 +824,22 @@ def calc_bootstrap(func, returns, *args, **kwargs):
     out = np.empty(n_samples)
 
     if len(args) > 0:
-        factor_returns = args.pop(0)
+        factor_returns = args[0]
+        args = args[1:]
+        # Ensure that factor_returns have same index
+        factor_returns = factor_returns[returns.index]
     else:
         factor_returns = None
 
     for i in range(n_samples):
         idx = np.random.randint(len(returns), size=len(returns))
+        returns_i = returns.iloc[idx].reset_index(drop=True)
         if factor_returns is not None:
-            out[i] = func(returns.iloc[idx],
-                          factor_returns.iloc[idx],
+            factor_returns_i = factor_returns.iloc[idx].reset_index(drop=True)
+            out[i] = func(returns_i, factor_returns_i,
                           *args, **kwargs)
         else:
-            out[i] = func(returns.iloc[idx],
+            out[i] = func(returns_i,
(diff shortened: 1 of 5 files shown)


twiecki:
image

twiecki:
image

twiecki:
@a-campbell Can you help reviewing this?

pyfolio/timeseries.py
```diff
@@ -752,6 +752,131 @@ def perf_stats(returns, factor_returns=None):
...
+    Returns
+    -------
+    numpy.ndarray
+        Bootstrapped sampling distribution of passed in func.
+    """
+
+    n_samples = kwargs.pop('n_samples', 1000)
+    out = np.empty(n_samples)
+
+    if len(args) > 0:
```
a-campbell: would it be clearer to pass factor returns in as a kwarg?

pyfolio/timeseries.py
```diff
@@ -752,6 +752,131 @@ def perf_stats(returns, factor_returns=None):
...
+        - Distributional statistics of bootstrapped sampling
+        distribution of performance metrics.
+        if return_stats is False:
+        - Bootstrap samples for each performance metric.
+    """
+    bootstrap_values = OrderedDict()
+
+    for stat_func in SIMPLE_STAT_FUNCS:
+        stat_name = stat_func.__name__
+        bootstrap_values[stat_name] = calc_bootstrap(stat_func,
```
a-campbell: You could go straight to the DataFrame here, but the dict -> df is fine

pyfolio/timeseries.py
```diff
@@ -752,6 +752,131 @@ def perf_stats(returns, factor_returns=None):
...
+        for stat_func in FACTOR_STAT_FUNCS:
+            stat_name = stat_func.__name__
+            bootstrap_values[stat_name] = calc_bootstrap(stat_func,
+                                                         returns,
+                                                         factor_returns)
+    bootstrap_values = pd.DataFrame(bootstrap_values)
+
+    if return_stats:
+        stats = bootstrap_values.apply(calc_distribution_stats)
+        return stats.T[['mean', '5%', '95%']]
```
a-campbell: Do we want to include median and/or IQR here? Would allow the user a way to asses the normality of the bootstrap results.

pyfolio/timeseries.py
```diff
@@ -752,6 +752,131 @@ def perf_stats(returns, factor_returns=None):
...
+    if len(args) > 0:
+        factor_returns = args[0]
+        args = args[1:]
+        # Ensure that factor_returns have same index
+        factor_returns = factor_returns[returns.index]
+    else:
+        factor_returns = None
+
+    for i in range(n_samples):
+        idx = np.random.randint(len(returns), size=len(returns))
```
a-campbell: I'm not 100% clear on the benefits/problems with sampling a random collection of daily returns vs. sampling a random window of daily returns. Is the assumption that daily returns are independent problematic?

a-campbell:
Looks great! Made a few comments. This is going to be a really interesting feature.

pyfolio/timeseries.py
twiecki: Neither am I but my hunch is that the difference won't be huge. There are various different ways to do a blocked bootstrap and we might want to add these. I'd lean towards sticking with the simplest one for now and if the need arises we can add different sampling schemes later.

pyfolio/timeseries.py
twiecki: Added median, IQR is similar enough to 5% and 95%.

PR #262, opened 2015-12-18
Refactor timeseries.py

### LINKED ISSUE #232: Remove scikits-learn and statsmodels dependencies
We use these for very simple functions we could easily call in scipy. That would make our requirements much slimmer.

twiecki:
Quite a large refactoring. 

* Put risk metrics to beginning of timeseries.py (unfortunately this, and the remove of old code mess up the diff).
* Add list of all risk metrics that can be computed just on returns and those which require factor returns.
* Removal of portfolio code that I think should go somewhere else and did not look like production code.
* Deprecate `out_of_sample_vs_in_sample_returns_kde()` and `stability_of_timeseries()`.
* Remove statsmodels and scikits-learn dependencies. Closes #232.

These changes are required for #261.

CC @llllllllll, @a-campbell, @justinlent

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import division
+from collections import OrderedDict
 
 import pandas as pd
 import numpy as np
@@ -510,13 +511,13 @@ def show_perf_stats(returns, factor_returns, live_start_date=None):
     perf_stats = np.round(timeseries.perf_stats(
         returns_backtest,
         factor_returns=factor_returns), 2)
-    perf_stats.columns = ['Backtest']
 
     if live_start_date is not None:
-        perf_stats = perf_stats.join(perf_stats_live,
-                                     how='inner')
-        perf_stats = perf_stats.join(perf_stats_all,
-                                     how='inner')
+        perf_stats = pd.DataFrame(OrderedDict([
+            ('Backtest', perf_stats),
+            ('Out of sample', perf_stats_live),
+            ('All history', perf_stats_all),
+        ]))
 
     print(perf_stats)
 
--- a/pyfolio/tests/test_timeseries.py
+++ b/pyfolio/tests/test_timeseries.py
@@ -289,7 +289,7 @@ class TestStats(TestCase):
             returns, rolling_sharpe_window).values.tolist()), expected)
 
     @parameterized.expand([
-        (simple_rets, 0.010766923838471554)
+        (simple_rets, 0.10376378866671222)
     ])
     def test_stability_of_timeseries(self, returns, expected):
         self.assertAlmostEqual(
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -697,40 +697,7 @@ def rolling_fama_french(returns, factor_returns=None,
                         rolling_window=rolling_window)
 
 
-def calc_alpha_beta(returns, factor_returns):
-    """Calculates both alpha and beta.
-
-    Parameters
-    ----------
-    returns : pd.Series
-        Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
-    factor_returns : pd.Series
-         Daily noncumulative returns of the factor to which beta is
-         computed. Usually a benchmark such as the market.
-         - This is in the same style as returns.
-
-    Returns
-    -------
-    float
-        Alpha.
-    float
-        Beta.
-
-"""
-
-    ret_index = returns.index
-    beta, alpha = sp.stats.linregress(factor_returns.loc[ret_index].values,
-                                      returns.values)[:2]
-
-    return alpha * APPROX_BDAYS_PER_YEAR, beta
-
-
-def perf_stats(
-        returns,
-        return_as_dict=False,
-        factor_returns=None,
-        period=DAILY):
+def perf_stats(returns, factor_returns=None):
     """Calculates various performance metrics of a strategy, for use in
     plotting.show_perf_stats.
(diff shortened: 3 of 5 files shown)


twiecki:
@llllllllll Would love your input on this. If you're swamped with other things no worries and I'll just go ahead without you.

pyfolio/timeseries.py
```diff
@@ -31,6 +28,9 @@
 from .interesting_periods import PERIODS
 
 
+#####################################
```
llllllllll: This indicates to me that we should split this into a risk.py module here or pull the non-risk metric code into another module.

twiecki: Yes, I actually started doing that already but backtracked because it wasn't clear that e.g. the rolling statistics shouldn't also go in there.

pyfolio/plotting.py
```diff
@@ -510,13 +511,13 @@ def show_perf_stats(returns, factor_returns, live_start_date=None):
...
         returns_backtest,
         factor_returns=factor_returns), 2)
-    perf_stats.columns = ['Backtest']
 
     if live_start_date is not None:
-        perf_stats = perf_stats.join(perf_stats_live,
-                                     how='inner')
-        perf_stats = perf_stats.join(perf_stats_all,
-                                     how='inner')
+        perf_stats = pd.DataFrame(OrderedDict([
```
llllllllll: This is cool, I didn't know dataframe respected ordered dict key ordering until now

pyfolio/tests/test_timeseries.py
```diff
@@ -289,7 +289,7 @@ def test_sharpe_2(self, returns, rolling_sharpe_window, expected):
             returns, rolling_sharpe_window).values.tolist()), expected)
 
     @parameterized.expand([
-        (simple_rets, 0.010766923838471554)
+        (simple_rets, 0.10376378866671222)
```
llllllllll: what caused this test to change, the only bug commit I see is the df join

twiecki: https://github.com/quantopian/pyfolio/commit/bb13b0a5c6166a230879d941389886aeed78bbe0. You're right that it should be a BUG commit as the original implementation wasn't correct.

pyfolio/timeseries.py
```diff
@@ -491,95 +493,129 @@ def stability_of_timeseries(returns):
...
-    kde_diff = sum(abs(scipy_kde_test - scipy_kde_train)) / \
-        (sum(scipy_kde_train) + sum(scipy_kde_test))
+    def cumulate_returns(x):
+        return cum_returns(x)[-1]
 
-    return kde_diff
+    if convert_to == WEEKLY:
+        return df_daily_rets.groupby(
+            [lambda x: x.year,
+             lambda x: x.month,
```
llllllllll: I don't think we need the month, the isocalendar week field is the weeknumber in the year,

twiecki: Neat, thanks.

pyfolio/timeseries.py
llllllllll: I would either split this or kill the comment. If the distinction is not clear now then I am not sure the comment helps.

pyfolio/timeseries.py
```diff
@@ -741,38 +738,22 @@ def perf_stats(
...
-    all_stats['sharpe_ratio'] = sharpe_ratio(
-        returns)
-    all_stats['calmar_ratio'] = calmar_ratio(returns, period=period)
-    all_stats['stability'] = stability_of_timeseries(returns)
-    all_stats['max_drawdown'] = max_drawdown(returns)
-    all_stats['omega_ratio'] = omega_ratio(returns)
-    all_stats['sortino_ratio'] = sortino_ratio(returns)
-    all_stats['skewness'] = stats.skew(returns)
-    all_stats['kurtosis'] = stats.kurtosis(returns)
+    stats = pd.Series()
```
llllllllll: This is considerably faster as:

```python
return pd.Series(OrderedDict(
    (stats_func.__name__, stats_func(returns)
    for stats_func in SIMPLE_STATS_FUNCS +
    (FACTOR_STAT_FUNCS if factor_returns else [])
))
```

Otherwise we resize the series everytime we add an element to the series.

pyfolio/timeseries.py
```diff
@@ -491,95 +493,129 @@ def stability_of_timeseries(returns):
...
-    df_cum_rets = cum_returns(returns, starting_value=100)
-    df_cum_rets_log = np.log10(df_cum_rets.values)
-    len_returns = df_cum_rets.size
+    return rhat
 
-    X = list(range(0, len_returns))
-    X = sm.add_constant(X)
 
-    model = sm.OLS(df_cum_rets_log, X).fit()
+SIMPLE_STAT_FUNCS = [
```
llllllllll: Do we intend on mutating these later? I would prefer an immutable structure for these module constants unless it is an intended feature that we can mutate them.

pyfolio/timeseries.py
twiecki: :+1:

pyfolio/timeseries.py
twiecki: Yeah, I have thought about it. A user might want to add to it actually so since we're not mutating this elsewhere I'd rather keep it mutable.

pyfolio/tests/test_timeseries.py
twiecki: Reworded that commit: https://github.com/quantopian/pyfolio/commit/ca8567e2ac2d9b1cae1c4242fdf571689d1416c8

twiecki:
@llllllllll Thanks for the great review. I addressed all issues.

PR #272, opened 2016-01-19
TST Install matplotlib-tests so that matplotlib can be unpegged. http…

twiecki:
…s://github.com/matplotlib/matplotlib/issues/5634

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests
   - source activate testenv
   - pip install nose_parameterized contextlib2 logbook==0.10.1
-  - pip install --no-deps git+https://github.com/quantopian/zipline@v0.8.3
+  #- pip install --no-deps git+https://github.com/quantopian/zipline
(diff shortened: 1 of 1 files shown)


twiecki:
Annoying voodo `ValueError: numpy.dtype has the wrong size, try recompiling` error again. https://github.com/conda/conda/issues/1701

ssanderson:
@twiecki that error means that you have a C Extension that was compiled against an older version of numpy's headers than you have installed.  This can happen, for example, if you compile and install module X with numpy 1.9, and then subsequently install numpy 1.10 without recompiling module X.  It's unlikely that this is conda's fault.

twiecki:
@ssanderson That's what I tried to look at but there shouldn't be any compilation going on as it's using conda to install everything.

ssanderson:
I think this might actually be a zipline bug: https://github.com/quantopian/zipline/blob/master/setup.py#L225-L227

ssanderson:
https://github.com/quantopian/zipline/issues/954

twiecki:
@ssanderson zipline is indeed the problem (it works when I remove it from travis).

ssanderson:
@twiecki you probably don't want to be unconditionally installing zipline master.  Are you depending on something ahead of the released 0.8.3 version?

twiecki:
No. We might actually be able to drop it and just include the returns and positions data of a backtest into pyfolio.

ssanderson:
> No. We might actually be able to drop it and just include the returns and positions data of a backtest into pyfolio.

I'd be nervous about doing this, since it will break if we change our output format.  I'd recommend just pip installing zipline==0.8.3 for now though.

twiecki:
0.8.3 seems to have the same problem.

PR #276, opened 2016-01-28
Tail ratio and common sense ratio

twiecki:
Felt inspired after reading https://www.quora.com/Under-what-market-conditions-does-mean-reversion-work-better-than-trend-following. Adds the two metrics proposed.


a-campbell:
LGTM. This SIMPLE_STAT_FUNCS pattern rocks.

PR #281, opened 2016-02-03
vectorize max drawdown and fix doc strings

devinstevenson:
* Vectorize max_drawdown for 4X speedup

* Fix typos in doc strings
model_returns_t and model_returns_normal had each other’s doc string

### DIFF
--- a/pyfolio/bayesian.py
+++ b/pyfolio/bayesian.py
@@ -106,11 +106,7 @@ def model_returns_t_alpha_beta(data, bmark, samples=2000):
 
 
 def model_returns_normal(data, samples=500):
-    """Run Bayesian model assuming returns are Student-T distributed.
-
-    Compared with the normal model, this model assumes returns be
-    T-distributed and thus has a 3rd parameter (nu) that controls the
-    mass in the tails.
+    """Run Bayesian model assuming returns are normally distributed.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


pyfolio/bayesian.py
```diff
@@ -149,7 +145,11 @@ def model_returns_normal(data, samples=500):
 
 
 def model_returns_t(data, samples=500):
-    """Run Bayesian model assuming returns are normally distributed.
+    """Run Bayesian model assuming returns are Student-T distributed.
+
+    Compared with the normal model, this model assumes returns are
+    T-distributed and thus have a 3rd parameter (nu) that controls the
+    mass in the tails.
```
twiecki: Good catch!

pyfolio/timeseries.py
```diff
@@ -76,18 +76,9 @@ def max_drawdown(returns):
...
-    peak = -99999
-    for value in df_cum_rets:
-        if (value > peak):
-            peak = value
-        else:
-            DD = (peak - value) / peak
-        if (DD > MDD):
-            MDD = DD
-    return -1 * MDD
+    return df_cum_rets.sub(cum_max_return).div(cum_max_return).min()
```
twiecki: Not only faster but dramatically simplified.

twiecki:
Thanks, that's great!

PR #287, opened 2016-03-09
Correct volatility match in plot_rolling_returns

wikiped:
Correct formula to calculate returns vs benchmark from:

```
returns = (returns / returns.std()) * bmark_vol
```

To:

```
returns = returns * (returns.std() / bmark_vol)
```

This will make a proper comparison of stock returns vs benchmark:

```
import pyfolio as pf
stock_rets = pf.utils.get_symbol_rets('FB')
bench_rets = pf.utils.get_symbol_rets('SPY')
```

Returns as they are:

```
pf.plotting.plot_rolling_returns(stock_rets, 
                                 factor_returns=bench_rets, 
                                 volatility_match=False)
```

ret1

Matched returns with original formula don't reflect the real dynamic (i.e `FB` doesn't cross `SPY`):

```
# returns = (returns / returns.std()) * bmark_vol                                 
pf.plotting.plot_rolling_returns(stock_rets, 
                                 factor_returns=bench_rets, 
                                 volatility_match=True)                                 
```

ret2

Matched returns with proposed formula has the same denominator as benchmark and therefore is relationship between the lines is not distorted:

```
# returns = returns * (returns.std() / bmark_vol)
pf.plotting.plot_rolling_returns(stock_rets, 
                                 factor_returns=bench_rets, 
                                 volatility_match=True)
```

ret3

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -632,7 +632,7 @@ def plot_rolling_returns(returns,
                          'factor_returns.')
     elif volatility_match and factor_returns is not None:
         bmark_vol = factor_returns.loc[returns.index].std()
-        returns = (returns / returns.std()) * bmark_vol
+        returns = returns * (returns.std() / bmark_vol)
 
     cum_rets = timeseries.cum_returns(returns, 1.0)
 


twiecki:
Not sure I know what you mean. The idea is to have both returns have the same volatility. So the fact that FB has higher cumulative returns isn't all that meaningful if it's 10x the volatility. Are you saying that's not what the computation does or that it should do something different?

wikiped:
I might have misread the original plot, but what I meant to illustrate is that current computation doesn't (in my opinion) do what it's supposed to do: which is to take away the 'scale of returns' effect and show how volatile are the stocks on a comparable base (the base is benchmark). The proposed change is supposed to make a comparison of 'apples-to-apples'. Perhaps if you could illustrate what the original chart (where `FB` doesn't cross benchmark) is saying it would solve misunderstanding. Thanks.

hkopp:
As far as I understand, the current computation shows how FB would have had evolved if it had had the same volatility as the benchmark. In my understanding it is not supposed to show how volatile the stocks are on a comparable base, as you said, but rather to equalize their volatility and compare their different returns, assuming the same vol.

On the other hand, I find it hard to follow your new proposal. The volatility of your adjusted returns are not the same as the volatility of the benchmark. What is the intuition behind your formula?

twiecki:
Thanks for the sanity check @hkopp.

PR #293, opened 2016-04-18
Refactor roundtrip

twiecki:
Complete rewrite of the round-trip analysis. Now does complete portfolio reconstruction based on transactions and matches orders even without crossing the zero-mark which should make it much more generally applicable. Also much better output.


twiecki:
CC @a-campbell

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+     ('Largest losing trade', 'min'),
+     ])
+
+DURATION_STATS = OrderedDict(
+    [('Avg duration', lambda x: x.mean()),
+     ('Median duration', lambda x: x.median()),
+     ('Avg # round_trips per day', lambda x: float(len(x)) /
+      (x.max() - x.min()).days),
+     ('Avg # round_trips per month', lambda x: float(len(x)) /
+      (((x.max() - x.min()).days) / 21)),
```
a-campbell: could use `utils.APPROX_BDAYS_PER_MONTH`

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+     ])
+
+
+def agg_all_long_short(round_trips, col, stats_dict):
+    stats_all = round_trips.assign(ones=1)\
```
a-campbell: Most examples of method chaining I've seen put parenthesis around the whole expression instead of using `\` 
ex: http://pandas.pydata.org/pandas-docs/stable/basics.html#tablewise-function-application

Not a big deal though.

twiecki: Oh that's why that didn't work.

pyfolio/round_trips.py
twiecki: +1

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+        for dt, t in trans_sym.iterrows():
+            if t.price < 0:
+                warnings.warn('Negative price detected, ignoring for'
+                              'round-trip.')
+                continue
+            signed_price = t.price * np.sign(t.amount)
+            abs_amount = int(abs(t.amount))
+            indiv_prices = [signed_price] * abs_amount
+            if (len(price_stack) == 0) or \
+               (copysign(1, price_stack[-1]) == copysign(1, t.amount)):
```
a-campbell: would `np.sign(price_stack[-1]) == np.sign(t.amount)` work here?

twiecki: That's what I had initially but there was a huge overhead with calling `numpy` on a scalar.

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+            .assign(date=portfolio_value.index)
+        roundtrips['date'] = roundtrips.close_dt.apply(lambda x:
+                                                       x.replace(hour=0,
+                                                                 minute=0,
+                                                                 second=0))
+
+        tmp = roundtrips.assign(date=roundtrips.close_dt)\
+                        .join(pv, on='date', lsuffix='_')
+
+        roundtrips['returns'] = tmp.pnl / tmp.portfolio_value
```
a-campbell: Do we want `tmp.pnl / tmp.porftolio_value.shift()`?

Alternatively, you could pass in `positions.sum(axis='columns') / (1+returns)` to set portfolio_value as the opening portfolio value. This way, the first day's return isn't NaN.

pyfolio/tears.py
```diff
@@ -153,13 +154,12 @@ def create_full_tear_sheet(returns,
     if returns.index[0] < benchmark_rets.index[0]:
         returns = returns[returns.index > benchmark_rets.index[0]]
 
-    if slippage is not None and transactions is not None:
+    if (unadjusted_returns is None) and (slippage is not None) and\
```
a-campbell: Is there a case where someone would be passing in both adjusted and unadjusted returns?

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+        price_stack = deque()
+        dt_stack = deque()
+        for dt, t in trans_sym.iterrows():
+            if t.price < 0:
+                warnings.warn('Negative price detected, ignoring for'
+                              'round-trip.')
+                continue
+            signed_price = t.price * np.sign(t.amount)
+            abs_amount = int(abs(t.amount))
+            indiv_prices = [signed_price] * abs_amount
```
a-campbell: could these operations be performed as array math outside of the loop?

pyfolio/round_trips.py
```diff
@@ -13,168 +13,253 @@
...
+            if (len(price_stack) == 0) or \
+               (copysign(1, price_stack[-1]) == copysign(1, t.amount)):
+                price_stack.extend(indiv_prices)
```
a-campbell: If `signed_price * abs_amount` is always a float, you could make `indiv_prices = signed_price * abs_amount` and do `price_stack.append(indiv_prices)` instead of `price_stack.extend(indiv_prices)`

twiecki: `indiv_prices = [t.signed_price] * t.abs_amount` creates a list of length `abs_amount`. So it's not a single value we're appending but a complete list of individual share prices.

pyfolio/tears.py
twiecki: Not sure how that ended up in here. I was thinking about the case where you want the full tearsheet on adjusted returns but also want the slippage sweep.

pyfolio/round_trips.py
twiecki: Went with your 2nd suggestion.

PR #294, opened 2016-04-18
Add log-scaling version of cumulative returns and raw returns plot to returns tear-sheet

twiecki:
image


a-campbell:
Looks like there's a flake8 error. Other than that, LGTM!

PR #298, opened 2016-05-13
Logo

richafrank:
Added pyfolio logo to README and docs

Fixed docs generation config, which had broken references


twiecki:
Awesome!

PR #301, opened 2016-05-19
Closes #296

### LINKED ISSUE #296: Add net exposure line to the Long/Short Exposure plot
Just subtract Short from Long and plot the difference as a timeseries.  
Also a couple of minor formatting changes.

Long exposure = _Green_
Short exposure = _Red_
Difference line = _Black dashed_

Make Long exposure line positive number, and Short exposure line all negative numbers:
(input from @Jstauth) rather than showing both as positive (abs) we should show the long as positive and short as negative, then we could overlay the net exposure as a line)

image

ahgnaw:


jbredeche:
Thanks for the PR!  It looks like the build is failing due to flake8 issues, can you run flake8 locally and sort those out?

twiecki:
@justinlent Can you review this?

justinlent:
@twiecki LGTM. Merge at your convenience

pyfolio/pos.py
```diff
@@ -57,13 +57,14 @@ def get_long_short_pos(positions):
 
     pos_wo_cash = positions.drop('cash', axis=1)
     longs = pos_wo_cash[pos_wo_cash > 0].sum(axis=1).fillna(0)
-    shorts = pos_wo_cash[pos_wo_cash < 0].abs().sum(axis=1).fillna(0)
+    shorts = pos_wo_cash[pos_wo_cash < 0].sum(axis=1).fillna(0)
```
twiecki: Won't removing `abs()` not change the results of the plot?

ahgnaw: Yes, it will. One of the issue tasks is to make the "shorts" line show up
as negative instead of positive.

"Make Long exposure line positive number, and Short exposure line all
negative numbers"

justinlent: LGTM

twiecki:
Thanks @ahgnaw!

pyfolio/pos.py
justinlent: @twiecki @ahgnaw  see line 62 below where cash is added in the `net_liquidation` value.  I believe this results in the cash taken in from shorts being double counted.  Thoughts?

a-campbell: if `shorts` is a negative number, net liquidation should be `longs + shorts + cash`

justinlent: from our offline chat:

The explicit calculation would be `net_liquidation = longs - shorts + (cash + shorts)

[12:39]  
which removes the short exposure from the cash component
which then begs the simplification to just `net_liquidation = longs  + cash`
will have to think about whether that makes sense though

PR #306, opened 2016-06-02
Overlay OOS datapoints in boxplot

ahgnaw:
Point markers are diamonds instead of crosses.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -1082,7 +1082,9 @@ def plot_sector_allocations(returns, sector_alloc, ax=None, **kwargs):
     return ax
 
 
-def plot_return_quantiles(returns, df_weekly, df_monthly, ax=None, **kwargs):
+def plot_return_quantiles(returns, df_weekly, df_monthly,
+                          live_start_date=None,
+                          ax=None, **kwargs):
     """Creates a box plot of daily, weekly, and monthly return
     distributions.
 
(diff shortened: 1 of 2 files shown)


twiecki:
Looking at this code I think it's time to refactor it. Specifically,
```
    df_weekly = timeseries.aggregate_returns(returns, 'weekly')
    df_monthly = timeseries.aggregate_returns(returns, 'monthly')
```

Should happen inside the plotting function and only `returns` be passed in. The slicing is a bit confusing and this refactor could remedy this.

ahgnaw:
@twiecki df_weekly and df_monthly are now only defined by the plotting functions that require them.

twiecki:
LGTM, thanks!

twiecki:
image

twiecki:
@ahgnaw just noticed that we should probably add a legend as well explaining the red dots.

ahgnaw:
image

Unfortunately matplotlib adds the line through the marker in the legend. What do you think of the text? Alternative: "Out-of-sample returns"

twiecki:
```python
import matplotlib as mpl
mpl.rcParams['legend.handlelength'] = 0
```
Perhaps using that instead in a `rc_context`: http://matplotlib.org/api/matplotlib_configuration_api.html#matplotlib.rc_context

Also, the monthly color red is a bit hard to read with the red markers.

twiecki:
I like out-of-sample data.

ahgnaw:
image

I'll try out a few different colors. The daily, weekly, and monthly markers could each be different colors.

twiecki:
I think we'd rather switch the red color of the last bar.

ahgnaw:
image

image

here's yellow or green. I have a slight preference for the green, but either works.

twiecki:
Sorry, I meant the actual fill color of the barplot, not the points.

ahgnaw:
Ah ok. I think this last. It's changed to a yellow that is consistent with the color palette. Thoughts?

image

twiecki:
Perfect.

PR #310, opened 2016-06-13
Multistrike cones

ahgnaw:
New plot draws a green cone of expectation around the out-of-sample returns. When returns drop outside of a cone, a new cone is redrawn. Up to 4 cones are drawn, green, yellow, orange, and red.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -23,6 +23,8 @@ import matplotlib
 import matplotlib.pyplot as plt
 from matplotlib.ticker import FuncFormatter
 import matplotlib.lines as mlines
+from matplotlib import figure
+from matplotlib.backends.backend_agg import FigureCanvasAgg
 
 from sklearn import preprocessing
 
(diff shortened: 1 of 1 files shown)


abhijeetkalyan:
Self-assigning, discussed with @twiecki

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+        the cone. If multiple values are passed, cone bounds will
+        be generated for each value.
+    random_seed : int
+        Seed for the pseudorandom number generator used by the pandas
+        sample method.
+
+
+    Returns
+    -------
+    ax : matplotlib.Axes
```
abhijeetkalyan: This confused me a bit, because it feels like we're returning both `ax` and `fig`, whereas what we're actually doing is returning axes if the caller passed them in, and if not, we're returning a new figure. Could we find a way to make that clearer in the docstring?

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
            xlim=(lower_plot, upper_plot), ylim=(0, y.max() + 1.))
 
     return ax
+
+
+def plot_cone(aggregate_returns, bounds, ax,
```
abhijeetkalyan: If `plot_cone` is a private method that's only going to be called by `plot_multistrike_cones`, maybe we should declare it as 'private' by prefacing it with an `_`?

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+    aggregate_returns : pandas.core.frame.DataFrame
+        Cumulative out-of-sample returns.
+    bounds : pandas.core.frame.DataFrame
+        Contains upper and lower cone boundaries. Column names are
+        strings corresponding to the number of standard devations
+        above (positive) or below (negative) the projected mean
+        cumulative returns.
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    cone_std : int, float, or list of int/float
```
abhijeetkalyan: Think this docstring needs to be updated - based on the code below and the default value above, it looks like `cone_std` needs to be an iterable of int/floats. (The for loop will `TypeError` with a single int/float)

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+    num_samples : int
+        Number of samples to draw from the in-sample daily returns.
+        Each sample will be an array with length num_days.
+        A higher number of samples will generate a more accurate
+        bootstrap cone.
+    name : str, optional
+        Plot title
+    ax : matplotlib.Axes, optional
+        Axes upon which to plot.
+    cone_std : int, float, or list of int/float
```
abhijeetkalyan: Same as above for `cone_std`, think this wants to say `iterable of int/float`

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+
+
+def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
+                           name=None, ax=None, cone_std=(1., 1.5, 2.),
+                           random_seed=None):
+    """
+    Plots the upper and lower bounds of an n standard deviation
+    cone of forecasted cumulative returns. This cone is non-parametric,
+    meaning it does not assume that returns are normally distributed. Redraws
+    a new cone when returns fall outside of last cone drawn
```
abhijeetkalyan: Missing a full stop at the end of the sentence :trollface:

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+    """
+    for std in cone_std:
+        ax.fill_between(aggregate_returns.index,
+                        bounds[float(std)].iloc[:len(aggregate_returns)],
+                        bounds[float(-std)].iloc[:len(aggregate_returns)],
+                        color=color, alpha=0.5)
+    return ax
+
+
+def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
```
abhijeetkalyan: As it currently stands, this function can plot up to 3 strikes, depending on how the out-of-sample returns behave - we should either update the function name and/or docstring to make it clear that it's a fixed number of strikes, or allow the user to pass in a parameter to determine how many strikes they care about. OK with either, maybe the former is better in terms of limiting scope?

abhijeetkalyan:
@ahgnaw can we add a screenshot of what the multiple cones look like to the PR description?

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,126 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+                                                random_seed=random_seed)
+    bounds.index = oos_returns.index
+    plot_cone(aggregate_returns, bounds, axes, cone_std=cone_std)
+
+    bounds_cur = bounds.copy()
+    cone_start = aggregate_returns.index[0]
+
+    aggregate_returns.plot(ax=axes,
+                           lw=3.,
+                           color='k',
```
abhijeetkalyan: This is a little(very, tbh!) nitpicky, but can we say `'black` instead of `k` here? It's a little more obvious what's going on with the explicit name instead of the shorthand.

pyfolio/plotting.py
```diff
@@ -1660,3 +1662,101 @@ def plot_prob_profit_trade(round_trips, ax=None):
...
+        Upper limit for number of cones drawn. Can be anything from 0 to 4.
+
+
+    Returns
+    -------
+    Returns are either an ax or fig option, but not both. If a
+    matplotlib.Axes instance is passed in as ax, then it will be modified
+    and returned. This allows for users to plot interactively in jupyter
+    notebook. When no ax object is passed in, a matplotlib.figure instance
+    is generated and returned.. This figure can then be used to save
```
abhijeetkalyan: Two full stops after `returned`

abhijeetkalyan:
Looks great! Just the one comment, then merge when travis is happy

ahgnaw:
num_strikes = 1
image
 num_strikes = 4

image

ahgnaw:
no_strikes = 0
image

no_strikes = 3
image

pyfolio/plotting.py
```diff
@@ -1726,17 +1688,26 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
-    cone_std : int, float, or list of int/float
+    cone_std : list of int/float
         Number of standard devations to use in the boundaries of
         the cone. If multiple values are passed, cone bounds will
         be generated for each value.
     random_seed : int
         Seed for the pseudorandom number generator used by the pandas
         sample method.
+    num_strikes : int
+        Upper limit for number of cones drawn. Can be anything from 0 to 3.
```
abhijeetkalyan: Minor nit on the docstring here - if num_strikes is 0, then we'll still have one cone drawn(even though we say in the docstring that the upper limit is 0). What do you think of rephrasing to something like:
```
num_strikes : int
  Number of times to re-draw a cone if we fall outside the bounds of the previous one. Can be any int from 0 to 3.
```

ahgnaw:
image

Multistrike cone chart of LinkedIn Stock with live start date of January 25, 2016 and num_strikes = 1

twiecki:
Awesome work @ahgnaw and @abhijeetkalyan!

PR #313, opened 2016-07-05
ENH: Added qrisk as dependency

ahgnaw:
Included qrisk as dependency. Functions within timeseries.py that are used to calculate risk metrics are wrappers for the equivalent function in qrisk. 

Tests within this repository were changed to reflect updated values.

### DIFF
--- a/pyfolio/tests/test_timeseries.py
+++ b/pyfolio/tests/test_timeseries.py
@@ -263,11 +263,12 @@ class TestAggregateReturns(TestCase):
         (simple_rets[:20], 'weekly', [0.3310000000000004, 0.0, 0.0])
     ])
     def test_aggregate_rets(self, returns, convert_to, expected):
-        self.assertEqual(
+        assert_almost_equal(
             timeseries.aggregate_returns(
                 returns,
-                convert_to).values.tolist(),
-            expected)
+                convert_to)['Agg'].tolist(),
+            expected,
+            DECIMAL_PLACES)
 
 
 class TestStats(TestCase):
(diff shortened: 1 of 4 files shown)


pyfolio/tests/test_timeseries.py
```diff
@@ -263,11 +263,12 @@ class TestAggregateReturns(TestCase):
         (simple_rets[:20], 'weekly', [0.3310000000000004, 0.0, 0.0])
     ])
     def test_aggregate_rets(self, returns, convert_to, expected):
-        self.assertEqual(
+        assert_almost_equal(
```
abhijeetkalyan: Why did this change? (Not that it's wrong, but context for the change)

pyfolio/timeseries.py
```diff
@@ -72,147 +74,122 @@ def max_drawdown(returns):
...
     ----------
     returns : pd.Series
         Periodic returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in cum_returns.
     period : str, optional
         - defines the periodicity of the 'returns' data for purposes of
         annualizing. Can be 'monthly', 'weekly', or 'daily'
         - defaults to 'daily'.
+    annualization : dict, optional
```
abhijeetkalyan: It looks like the default is `None`, not the dictionary we passed in here?

twiecki:
I don't think this is the right way, we should completely remove the overlapping functions and tests and call them in `qrisk` directly.

ahgnaw:
@twiecki I did it this way to maintain backwards compatibility. Eventually backwards compatibility will be broken, we should decide when.

ahgnaw:
@twiecki What about notifying users that these functions are being deprecated? We can break backwards compatibility at a major release.

twiecki:
@ahgnaw I like that.

PR #314, opened 2016-07-06
Add qrisk

ahgnaw:
Use qrisk for risk calculations. To preserve backwards compatibility, risk calculation functions are wrappers for corresponding qrisk functions. Risk calculation throughout the project updated to use qrisk instead of internal definition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
-  - pip install nose_parameterized contextlib2 logbook==0.10.1
+  - pip install nose_parameterized contextlib2 logbook==0.10.1 qrisk
   #- pip install --no-deps git+https://github.com/quantopian/zipline
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
   - pip install --no-deps git+https://github.com/Theano/Theano.git@rel-0.8.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
-  - pip install nose_parameterized
+  - pip install nose_parameterized qrisk>=0.1.4
   #- pip install --no-deps git+https://github.com/quantopian/zipline
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
   - pip install --no-deps git+https://github.com/Theano/Theano.git@rel-0.8.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
-  - pip install nose_parameterized contextlib2 logbook==0.10.1 qrisk>=0.1.3
+  - pip install nose_parameterized
   #- pip install --no-deps git+https://github.com/quantopian/zipline
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
   - pip install --no-deps git+https://github.com/Theano/Theano.git@rel-0.8.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
-  - pip install nose_parameterized contextlib2 logbook==0.10.1 qrisk
+  - pip install nose_parameterized contextlib2 logbook==0.10.1 qrisk>=0.1.3
   #- pip install --no-deps git+https://github.com/quantopian/zipline
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
   - pip install --no-deps git+https://github.com/Theano/Theano.git@rel-0.8.1
--- a/pyfolio/deprecate.py
+++ b/pyfolio/deprecate.py
@@ -1,0 +1,45 @@
+"""Utilities for marking deprecated functions."""
+# Copyright 2016 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+from functools import wraps
+
+
+def deprecated(msg=None, stacklevel=2):
+    """
+    Used to mark a function as deprecated.
+    Parameters
+    ----------
+    msg : str
+        The message to display in the deprecation warning.
+    stacklevel : int
+        How far up the stack the warning needs to go, before
+        showing the relevant calling lines.
+    Usage
+    -----
+    @deprecated(msg='function_a is deprecated! Use function_b instead.')
+    def function_a(*args, **kwargs):
+    """
+    def deprecated_dec(fn):
+        @wraps(fn)
+        def wrapper(*args, **kwargs):
+            warnings.warn(
+                msg or "Function %s is deprecated." % fn.__name__,
+                category=DeprecationWarning,
+                stacklevel=stacklevel
+            )
+            return fn(*args, **kwargs)
+        return wrapper
+    return deprecated_dec
--- a/pyfolio/tests/test_timeseries.py
+++ b/pyfolio/tests/test_timeseries.py
@@ -6,7 +6,6 @@ from numpy.testing import assert_allclose, assert_almost_equal
 
 import numpy as np
 import pandas as pd
-import pandas.util.testing as pdt
 
 from .. import timeseries
 from .. import utils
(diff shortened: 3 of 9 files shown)


pyfolio/timeseries.py
```diff
@@ -72,57 +74,29 @@ def max_drawdown(returns):
...
     returns : pd.Series
         Periodic returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
     period : str, optional
-        - defines the periodicity of the 'returns' data for purposes of
-        annualizing. Can be 'monthly', 'weekly', or 'daily'
-        - defaults to 'daily'.
+        Defines the periodicity of the 'returns' data for purposes of
+        annualizing. Can be 'monthly', 'weekly', or 'daily'.
```
abhijeetkalyan: Looks like the bit about the defaults is still true?

pyfolio/timeseries.py
```diff
@@ -208,11 +155,13 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
     annual_return_threshold : float, optional
-        Threshold over which to consider positive vs negative
-        returns. For the ratio, it will be converted to a daily return
-        and compared to returns.
+        Minimum acceptance return of the investor. Threshold over which to
```
abhijeetkalyan: Minor, but here we say `minimum acceptance return`, and below we say `minimum acceptable return`

pyfolio/timeseries.py
```diff
@@ -247,13 +185,12 @@ def sortino_ratio(returns, required_return=0, period=DAILY):
...
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
     required_return: float / series
         minimum acceptable return
     period : str, optional
-        - defines the periodicity of the 'returns' data for purposes of
-        annualizing. Can be 'monthly', 'weekly', or 'daily'
-        - defaults to 'daily'.
+        Defines the periodicity of the 'returns' data for purposes of
+        annualizing. Can be 'monthly', 'weekly', or 'daily'.
```
abhijeetkalyan: Same issue with the defaults

pyfolio/tests/test_timeseries.py
```diff
@@ -311,12 +311,13 @@ class TestStats(TestCase):
         (simple_month_rets, utils.MONTHLY, 0.006885932704891129)
     ])
     def test_annual_ret(self, returns, period, expected):
-        self.assertEqual(
+        assert_almost_equal(
```
abhijeetkalyan: Why did these change from `assertEqual` to `assert_almost_equal`?

pyfolio/timeseries.py
```diff
@@ -72,57 +74,29 @@ def max_drawdown(returns):
...
 def annual_return(returns, period=DAILY):
-    """Determines the annual returns of a strategy.
+    """Determines the mean annual growth rate of returns.
 
     Parameters
     ----------
     returns : pd.Series
         Periodic returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
```
abhijeetkalyan: Typo `timseries`

pyfolio/timeseries.py
```diff
@@ -208,11 +155,13 @@ def omega_ratio(returns, annual_return_threshhold=0.0):
...
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
     annual_return_threshold : float, optional
-        Threshold over which to consider positive vs negative
-        returns. For the ratio, it will be converted to a daily return
-        and compared to returns.
+        Minimum acceptance return of the investor. Threshold over which to
+        consider positive vs negative returns. It will be converted to a
+        value appropriate for the period of the returns. E.g. An annual minimum
```
abhijeetkalyan: Do you think it's worth explaining this a little bit more? I didn't get how we go from 100 to 0.018

.travis.yml
```diff
@@ -26,7 +26,7 @@ before_install:
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy statsmodels flake8 scikit-learn seaborn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
-  - pip install nose_parameterized contextlib2 logbook==0.10.1
+  - pip install nose_parameterized contextlib2 logbook==0.10.1 qrisk>=0.1.3
```
richafrank: Why does this differ from setup.py?  Instead of specifying it twice, could we pip install pyfolio itself here, which will look at any dependencies?  I notice that logbook and contextlib2 are not install requirements - why do we install them here?

twiecki: Left-overs from zipline, can be removed.

twiecki:
Also needs deprecation warnings when the old functions are called.

pyfolio/tests/test_timeseries.py
ahgnaw: This test was failing on my computer `AssertionError: 0.15500998835658053 != 0.15500998835658075` it was an insignificant enough difference that I didn't see it as a bigger issue.

pyfolio/timeseries.py
ahgnaw: The default options haven't changed.

pyfolio/timeseries.py
ahgnaw: Good catch

pyfolio/tests/test_timeseries.py
abhijeetkalyan: Yup, I get the logic of why we want to use `assert_almost_equal` in the general case (even 0.1 + 0.2 has floating-point issues). What I meant to ask was why it changed between this branch and master - was the test failing on your computer on master as well?

abhijeetkalyan:
For deprecation, we could just use https://github.com/quantopian/zipline/blob/master/zipline/utils/deprecate.py (the deprecation code will be duplicated, but probably not worth our time to set up a shared utilities repo for the sake of one file ;)
and then decorate the appropriate functions with `@deprecated`

pyfolio/timeseries.py
abhijeetkalyan: Yup, but in the docstring we removed the 'defaults to daily' line. Should we leave it in, since the defaults haven't changed?

pyfolio/timeseries.py
```diff
@@ -169,36 +129,23 @@ def calmar_ratio(returns, period=DAILY):
     ----------
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
-         - See full explanation in tears.create_full_tear_sheet.
+        - See full explanation in :func:`~pyfolio.timseries.cum_returns`.
```
abhijeetkalyan: several `timseries` -> `timeseries`, might be worth a find and replace :trollface:

twiecki:
yeah can just copy&paste that.

pyfolio/timeseries.py
```diff
@@ -410,23 +312,19 @@ def alpha_beta(returns, factor_returns):
...
-                                      returns.values)[:2]
+    """
 
-    return alpha * APPROX_BDAYS_PER_YEAR, beta
+    return qrisk.alpha_beta(returns, factor_returns=factor_returns)
 
 
 def alpha(returns, factor_returns):
-    """Calculates annualized alpha.
+    """Calculates alpha.
```
abhijeetkalyan: Is it no longer annualized?

pyfolio/timeseries.py
```diff
@@ -600,31 +493,19 @@ def cum_returns(returns, starting_value=None):
...
     """
 
-    # df_price.pct_change() adds a nan in first position, we can use
-    # that to have cum_returns start at the origin so that
-    # df_cum.iloc[0] == starting_value
-    # Note that we can't add that ourselves as we don't know which dt
-    # to use.
-    if pd.isnull(returns.iloc[0]):
-        returns.iloc[0] = 0.
+    starting = 0 if starting_value is None else starting_value
```
abhijeetkalyan: Can we simplify this by just passing `starting_value=0` as the default argument, and removing this line? IMO this pattern is generally more useful if the default argument is mutable, eg if you had `starting_value=[]`, but in this case we can just pass `starting_value=0` as the default.

abhijeetkalyan:
Looks good, had some comments

pyfolio/timeseries.py
ahgnaw: I'll leave it in. The default is specified in the function definition so I removed it from the docstring.

pyfolio/tests/test_timeseries.py
ahgnaw: The test test passes on master on my computer. The issue came up when switched over to qrisk.

.travis.yml
ahgnaw: That is an oversight. Moved over to `etc/requirements.txt`

pyfolio/timeseries.py
ahgnaw: It was definitely caused by find all and replace.

pyfolio/tests/test_timeseries.py
ahgnaw: In `cum_returns` there was a slight change in the calculation. `np.log(1 + returns)` was changed to `np.log1p(returns)` and that is causing the slight difference in the values for 'cum_returns`, `annual_return`, and `aggregate_returns`.

twiecki: Then we should just update the expected value.

pyfolio/timeseries.py
ahgnaw: No, it is not.

twiecki: Was it before? Shouldn't it be?

pyfolio/timeseries.py
ahgnaw: Should we change the function api in pyfolio? I didn't want to change in in case `None` is being passed in explicitly somewhere.

pyfolio/timeseries.py
ahgnaw: It previously assumed daily returns and annualized alpha. It's unclear to me what it should be.

@Jstauth @justinlent Is it conventional to annualize alpha?

ahgnaw: @twiecki I asked Jess and it should be, I'll change it in qrisk and here.

pyfolio/deprecate.py
```diff
@@ -0,0 +1,45 @@
+"""Utilities for marking deprecated functions."""
+# Copyright 2016 Quantopian, Inc.
```
abhijeetkalyan: I think we can kill this copyright comment here actually, from what I understand we only need to have this once in the README. @richafrank @twiecki any info on this?

ahgnaw:
Just made a new commit that adds deprecation warnings to functions defined in timeseries.py, uses qrisk functions whenever risk calculations are used within the project, and moves risk function testing out of this project. Testing for functions defined in qrisk should happen in that repo instead of having duplicates here.

pyfolio/deprecate.py
twiecki: I don't know the answer to that. I think we were told to do it like that from our OSS expert advisor.

richafrank: My non-expert reading of [How to apply this license] (http://choosealicense.com/licenses/apache-2.0/) is that we should include it here:

> The Apache Foundation recommends taking the additional step of adding a boilerplate notice to the header of each source file. You can find the notice at the very end of the license in the appendix.

ahgnaw: @richafrank Does this all mean each file in qrisk should have a copyright notice?

PR #320, opened 2016-08-12
ENH: Use empyrical instead of qrisk

ahgnaw:
The name of the risk library changed from qrisk to empyrical. This affected pyfolio in several places. 

An update in empyrical also requires function inputs (i.e. 'returns' and 'factor_returns') to have identical indexes. As a result, a test required an adjustment.

Example data has been rebuilt to reflect updated metrics. The affected columns are: treasury_period_return, sortino, beta, benchmark_volatility, benchmark_period_return, alpha, algorithm_period_return, algo_volatility.


abhijeetkalyan:
> An update in empyrical also requires function inputs (i.e. 'returns' and 'factor_returns') to have identical indexes.

Looks like the same update also led to changes in `plotting.py` like the one below?
```python
factor_returns=factor_returns[returns_backtest.index])
```

PR #327, opened 2016-09-08
MAINT: Removed calls to cum_returns just to get the result's index

richafrank:
when we can get it from the args to cum_returns


twiecki:
LGTM, thanks!

PR #328, opened 2016-09-18
BLD: pyfolio can use newer empyrical

richafrank:
Also don't need special install for empyrical on travis

We don't have any pyfolio C extensions, but when we do, pip installing pyfolio should include that.


twiecki:
:+1:

PR #329, opened 2016-09-21
ENH: Use all available return data for summary metrics

ahgnaw:
Keep all available returns data until risk metrics are calculated in `create_full_tear_sheet`.


richafrank:
@twiecki would you take a look at this?

PR #330, opened 2016-09-22
Update requirements

ssanderson:


pyfolio/ipycompat.py
```diff
@@ -0,0 +1,19 @@
+import IPython
```
ahgnaw: Could we put this in utils.py?

ahgnaw:
Pending tests, LGTM

pyfolio/ipycompat.py
ssanderson: Talked about this in person.  If we add a `utils` directory, I could see moving this there.  I don't think this belongs in single `utils` file.

twiecki:
Thanks!

PR #331, opened 2016-09-26
BUG: PyFolio depends on seaborn>=0.7.0 for swarmplot.

ssanderson:
PyFolio claims to depend on `seaborn>=0.6`, but it uses `seaborn.swarmplot`, which was added in 0.7.


dunster:
https://github.com/quantopian/qexec/issues/11424

twiecki:
Thanks!

PR #332, opened 2016-09-29
Portfolio multistrike cones

ahgnaw:
A portfolio cone prediction changes as new funds are added. Previous method treated portfolio returns without regard to when new funds were turned on.

This pr also fixes the issue that some conecharts were not generated.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -1659,8 +1659,8 @@ def plot_prob_profit_trade(round_trips, ax=None):
     return ax
 
 
-def _plot_cones(name, bounds, oos_returns, num_samples=1000, ax=None,
-                cone_std=(1., 1.5, 2.), random_seed=None, num_strikes=3):
+def plot_cones(name, bounds, oos_returns, num_samples=1000, ax=None,
+               cone_std=(1., 1.5, 2.), random_seed=None, num_strikes=3):
     """
     Plots the upper and lower bounds of an n standard deviation
     cone of forecasted cumulative returns. Redraws a new cone when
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -29,7 +29,6 @@ install:
   - pip install nose_parameterized
   #- pip install --no-deps git+https://github.com/quantopian/zipline
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then conda install --yes mock enum34; fi
-  - pip install --no-deps git+https://github.com/pymc-devs/pymc3.git
```
abhijeetkalyan: Why this change?

pyfolio/plotting.py
```diff
@@ -1659,28 +1659,29 @@ def plot_prob_profit_trade(round_trips, ax=None):
     return ax
 
 
-def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
-                           name=None, ax=None, cone_std=(1., 1.5, 2.),
-                           random_seed=None, num_strikes=0):
+def plot_cones(name, bounds, oos_returns, num_samples=1000, ax=None,
```
abhijeetkalyan: Should the name explain that this cone still plots multiple strikes?

pyfolio/plotting.py
```diff
@@ -1734,27 +1728,181 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
-            (returns.iloc[-1] - 1) * 100)
-    )
+    label = 'Cumulative returns = {:.2f}%'.format((returns.iloc[-1] - 1) * 100)
+    axes.plot(returns.index, returns.values, color='k', lw=3.,
+              label=label)
+
     if name is not None:
         axes.set_title(name)
-    axes.axhline(1, color='black', alpha=0.2)
+    axes.axhline(1, color='k', alpha=0.2)
```
abhijeetkalyan: Not a huge deal, but 'black' is a little clearer than 'k' - with the latter, I have to know matplotlib's colour <-> string mapping

pyfolio/plotting.py
abhijeetkalyan: If this only meant to be called internally, we can make it a private method.

pyfolio/plotting.py
```diff
@@ -1734,27 +1728,181 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
+    cone of forecasted cumulative returns. Redraws a new cone when returns
+    fall outside of last cone drawn.
+
+    Parameters
+    ----------
+    live_all : pandas.core.panel.Panel
+        Account names are items, the major axis is
+        pandas.tseries.index.DatetimeIndex, and the minor axis contains a
+        'returns' column. Each 'returns' column contains non-cumulative
+        in-sample returns.
```
abhijeetkalyan: Looks like this should be out of sample? Since we generate `oos_returns` from `live_all`

.travis.yml
ahgnaw: pymc3 is installed by `pip install .[bayesian]`. There's no need to install it twice. This dev version was also missing `pymc3.sqrt` and causing some tests to fail.

pyfolio/plotting.py
ahgnaw: I prefer `black` too, I just wanted to be consistent and changed this to match an earlier line. I'll change it to back to `black`.

pyfolio/plotting.py
ahgnaw: Good catch, thanks!

pyfolio/plotting.py
```diff
@@ -1758,3 +1750,159 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
+
+def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
+                           name=None, ax=None, cone_std=(1., 1.5, 2.),
+                           random_seed=None, num_strikes=0):
+    """
+    Plots the upper and lower bounds of an n standard deviation
+    cone of forecasted cumulative returns. This cone is non-parametric,
+    meaning it does not assume that returns are normally distributed. Redraws
+    a new cone when returns fall outside of last cone drawn.
+    Parameters
```
twiecki: Add new-line before.

pyfolio/plotting.py
```diff
@@ -1758,3 +1750,159 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
+        be generated for each value.
+    random_seed : int
+        Seed for the pseudorandom number generator used by the pandas
+        sample method.
+    num_strikes : int
+        Upper limit for number of cones drawn. Can be anything from 0 to 3.
+
+    Returns
+    -------
+    Returns are either an ax or fig option, but not both. If a
```
twiecki: -are

pyfolio/plotting.py
```diff
@@ -1758,3 +1750,159 @@ def plot_multistrike_cones(is_returns, oos_returns, num_samples=1000,
...
+    matplotlib.Axes instance is passed in as ax, then it will be modified
+    and returned. This allows for users to plot interactively in jupyter
+    notebook. When no ax object is passed in, a matplotlib.figure instance
+    is generated and returned. This figure can then be used to save
+    the plot as an image without viewing it.
```
twiecki: Hm, that's pretty odd. Can't we just always return an `ax`?

twiecki:
I'm not sure if `plot_portfolio_cones` shouldn't live in `qids` as it's a not really useful to single portfolio analysis.

Otherwise, is there perhaps an example plot you can post that shows how it's working?

ahgnaw:
@twiecki I wasn't sure where it should go. Moving it to qids makes sense.

pyfolio/plotting.py
ahgnaw: We store save the `figure` object as a png to be viewed in qdash.

ahgnaw:
@twiecki I moved the method over to qids

pyfolio/plotting.py
abhijeetkalyan: More generally, this allows us to work both in side-effectful environments, like a notebook, as well as environments where we don't want to have side effects, like in downstream apps.

pyfolio/timeseries.py
```diff
@@ -1040,24 +1112,17 @@ def forecast_cone_bootstrap(is_returns, num_days, cone_std=(1., 1.5, 2.),
...
+        is_returns=is_returns,
+        num_days=num_days,
+        starting_value=starting_value,
+        num_samples=num_samples,
+        random_seed=random_seed
+    )
+
+    cone_bounds = summarize_paths(
+        samples=samples,
+        cone_std=cone_std)
```
abhijeetkalyan: Nitpicky, but can we have the parenthesis on the next line

PR #334, opened 2016-10-03
Cum returns final

ahgnaw:
https://github.com/quantopian/qexec/issues/8187

In the backtester we display 'Total Returns' which is the last value of the cumulative returns. In pyfolio, calling `perf_stats(returns)` displays annual return, but not the total return.


twiecki:
LGTM.

twiecki:
:+1:

PR #339, opened 2016-10-13
BUG Disable round_trips per day and round_trips per month as they are buggy.

twiecki:
This only disables two measures which are buggy.


ahgnaw:
Aside from the flake8 issue, lgtm

PR #351, opened 2016-11-17
Detect intraday strategies and remove gross_lev

gusgordon:
Detect intraday strategies using the new `utils.detect_intraday` function, and resample the positions for intraday algorithms using the new `utils.estimate_intraday` function. This is applied automatically in tear sheets, but the user can override the detection. If users want to run a single plotting function with intraday-corrected returns, they can call `utils.estimate_intraday`, then pass the corrected positions to the plotting function.

Additionally, remove `gross_lev` as an argument. When needed, gross leverage will be calculated with `timeseries.gross_lev`, which is passed `positions` instead.


gusgordon:
@twiecki Can you review? Thanks.

pyfolio/tears.py
```diff
@@ -58,12 +58,12 @@ def create_full_tear_sheet(returns,
                            transactions=None,
                            market_data=None,
                            benchmark_rets=None,
-                           gross_lev=None,
                            slippage=None,
                            live_start_date=None,
                            sector_mappings=None,
                            bayesian=False,
                            round_trips=False,
+                           estimate_intraday=None,
```
twiecki: If bool should default to `True` or `False`.

pyfolio/tears.py
```diff
@@ -160,10 +157,24 @@ def create_full_tear_sheet(returns,
         unadjusted_returns = returns.copy()
         returns = txn.adjust_returns_for_slippage(returns, turnover, slippage)
 
+    if positions is not None and transactions is not None:
+        if (estimate_intraday is None and
+                utils.detect_intraday(positions, transactions)):
+            warnings.warn('Detected intraday strategy. Set ' +
```
twiecki: Would add that we will infer positions from transactions.

pyfolio/tears.py
```diff
@@ -160,10 +157,24 @@ def create_full_tear_sheet(returns,
...
+            positions = utils.estimate_intraday(returns, positions,
+                                                transactions)
+
+    elif estimate_intraday is True:
+        raise ValueError('Positions and transactions needed' +
+                         'to estimate intraday')
+
     create_returns_tear_sheet(
         returns,
+        positions=positions,
```
twiecki: Why does `returns` take `positions` now?

pyfolio/tears.py
```diff
@@ -503,10 +530,22 @@ def create_txn_tear_sheet(returns, positions, transactions,
...
          - See pyfolio.plotting.plot_swippage_sleep and
            pyfolio.plotting.plot_slippage_sensitivity
+    estimate_intraday: boolean, optional
+        Approximate returns for intraday strategies.
+        See description in create_full_tear_sheet.
     return_fig : boolean, optional
         If True, returns the figure that was plotted on.
     """
 
+    if (estimate_intraday is None and
```
twiecki: If we have more or less identical code 3 times should refactor that into a function too.

pyfolio/tears.py
```diff
@@ -452,22 +484,16 @@ def create_position_tear_sheet(returns, positions, gross_lev=None,
...
     plotting.plot_holdings(returns, positions_alloc, ax=ax_holdings)
 
-    last_pos = 4
-    if gross_lev is not None:
-        ax_gross_leverage = plt.subplot(gs[last_pos, :],
-                                        sharex=ax_exposures)
-        plotting.plot_gross_leverage(returns, gross_lev,
-                                     ax=ax_gross_leverage)
-        last_pos += 1
+    plotting.plot_gross_leverage(returns, positions,
```
twiecki: Nice simplification.

pyfolio/tears.py
twiecki: I see, `None` means to infer. This appears a bit brittle. Instead I would do `estimate_intraday='infer'` to make this explicit.

pyfolio/tests/test_tears.py
```diff
@@ -39,11 +39,6 @@ class PositionsTestCase(TestCase):
             __location__ + '/test_data/test_returns.csv.gz'),
         index_col=0, parse_dates=True)
     test_returns = to_series(to_utc(test_returns))
-    test_gross_lev = read_csv(
-        gzip.open(
-            __location__ + '/test_data/test_gross_lev.csv.gz'),
```
twiecki: Great to remove it here, I would, however, add a comparison between inferred gross_lev and what we load from the file.

pyfolio/timeseries.py
```diff
@@ -630,8 +630,28 @@ def rolling_fama_french(returns, factor_returns=None,
...
+        Daily net position values.
+         - See full explanation in tears.create_full_tear_sheet.
+
+    Returns
+    -------
+    pd.Series
+        Gross leverage.
+    """
+    exposure = positions.drop('cash', axis=1).abs().sum(axis=1)
+    return exposure/positions.sum(axis=1)
```
twiecki: pep8 needs spaces around math operators.

twiecki: Thinking about an example:
100% long, 100% short. lev should be 2. Is that the outcome here? Seems like positions.sum() is 0 in that case which would cause infinite leverage.

pyfolio/utils.py
```diff
@@ -541,3 +537,120 @@ def standardize_data(x):
...
+    Returns
+    -------
+    returns : boolean
+        True if an intraday strategy is detected.
+    """
+
+    daily_txn = transactions.copy()
+    daily_txn.index = daily_txn.index.date
+    txn_count = daily_txn.groupby(level=0).symbol.nunique().sum()
+    daily_pos = positions.drop('cash', axis=1).replace(0, np.nan)
```
twiecki: Good idea.

pyfolio/timeseries.py
twiecki: This function definitely needs tests.

pyfolio/utils.py
```diff
@@ -541,3 +537,120 @@ def standardize_data(x):
     """
 
     return (x - np.mean(x)) / np.std(x)
+
+
+def detect_intraday(positions, transactions, threshold=0.25):
```
twiecki: These two functions also need tests.

pyfolio/tests/test_tears.py
gusgordon: That's a good idea.

pyfolio/timeseries.py
gusgordon: In that case, positions.sum() would also add cash, so it wouldn't be zero. There could be other cases where there's failure's though. I'll make this more robust and add some tests.

pyfolio/plotting.py
```diff
@@ -354,7 +354,7 @@ def plot_holdings(returns, positions, legend_loc='best', ax=None, **kwargs):
         ax = plt.gca()
 
     positions = positions.copy().drop('cash', axis='columns')
-    df_holdings = positions.apply(lambda x: np.sum(x != 0), axis='columns')
+    df_holdings = positions.replace(0, np.nan).count(axis=1)
```
twiecki: `.fillna(0)`

pyfolio/tears.py
```diff
@@ -722,7 +740,8 @@ def create_capacity_tear_sheet(returns, positions, transactions,
                                liquidation_daily_vol_limit=0.2,
                                trade_daily_vol_limit=0.05,
                                last_n_days=utils.APPROX_BDAYS_PER_MONTH * 6,
-                               days_to_liquidate_limit=1):
+                               days_to_liquidate_limit=1,
+                               estimate_intraday='infer'):
```
twiecki: why not forward the kwarg?

PR #353, opened 2016-11-23
Exposure bug fix

gusgordon:
Fixes issue with exposure, removes `get_long_short_pos`.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -936,18 +936,18 @@ def plot_exposures(returns, positions, ax=None, **kwargs):
 
     pos_no_cash = positions.copy().drop('cash', axis=1)
     l_exp = pos_no_cash[pos_no_cash > 0].sum(axis=1) / positions.sum(axis=1)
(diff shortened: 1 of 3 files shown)


pyfolio/plotting.py
```diff
@@ -934,13 +934,26 @@ def plot_exposures(returns, positions_alloc, ax=None, **kwargs):
     if ax is None:
         ax = plt.gca()
 
-    df_long_short = pos.get_long_short_pos(positions_alloc)
-    df_long_short.plot(
-        kind='line', style=['-g', '-r', '--k'], alpha=1.0,
-        ax=ax, **kwargs)
+    pos_no_cash = positions.copy().drop('cash', axis=1)
```
twiecki: I'm pretty sure `.drop()` already creates a copy.

twiecki:
Actually, I think we should keep `get_long_short_pos`, users might depend on it.

gusgordon:
Sure, I'll add that back Monday.

PR #358, opened 2016-12-13
ENH: Format zipline assets before displaying

gusgordon:
Adds a new function, `utils.format_asset`, which will format zipline assets if they are about to be displayed. The formatting is done only just before the display.

Also bumps the latest version of PyMC3 for tests.

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -135,10 +135,9 @@ def _1_bday_ago():
 
 def format_asset(asset):
     """
-    If zipline asset objects are used, we want to print it out prettily
-    within the tear sheet. We wrap the whole call in a try, except
-    so we are also compatible with older versions of zipline. This function
-    should only be applied directly before displaying.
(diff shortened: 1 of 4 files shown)


pyfolio/utils.py
```diff
@@ -133,6 +133,24 @@ def _1_bday_ago():
...
+    """
+    If zipline asset objects are used, we want to print it out prettily
+    within the tear sheet. We wrap the whole call in a try, except
+    so we are also compatible with older versions of zipline. This function
+    should only be applied directly before displaying.
+    """
+
+    try:
+        import zipline.assets._assets as zipline_assets
+        if isinstance(asset, (zipline_assets.Equity, zipline_assets.Future)):
```
richafrank: Good use case for the base class `zipline.assets.Asset`?

gusgordon: Yes - thanks!

PR #362, opened 2017-02-07
Use latest PyMC3 for tests and update whatsnew for 0.7

gusgordon:


WHATSNEW.md
```diff
@@ -2,7 +2,27 @@
 
 These are new features and improvements of note in each release.
 
-## v0.6.0 (Oct, 17, 2016)
+## v0.7.0 (Jan 28th, 2017)
+
+This is a major release from `0.6.0`, and all users are recommended to upgrade.
+
+### New features
+
```
twiecki: Should mention new empyrical split.

gusgordon: Those functions were broken out for v0.6, I think. Or are you talking about `cum_returns` for the cones?

twiecki: Disregard, I saw we already included it in the release notes for 0.6.

Issue #363, opened 2017-02-14
Add longest and shortest holding times to round trips table


gusgordon:
It would be useful if this table also had minimum and maximum holding duration:
![pasted image at 2017_02_14 11_01 am](https://cloud.githubusercontent.com/assets/1934957/22937259/115e0988-f2a6-11e6-89b7-ec210c181e7a.png)

PR #367, opened 2017-03-02
Add min and max holding time to round trips

gusgordon:
Closes https://github.com/quantopian/pyfolio/issues/363
screen shot 2017-03-02 at 2 20 15 pm

### DIFF
--- a/pyfolio/round_trips.py
+++ b/pyfolio/round_trips.py
@@ -59,6 +59,8 @@ RETURN_STATS = OrderedDict(
 DURATION_STATS = OrderedDict(
     [('Avg duration', lambda x: x.mean()),
      ('Median duration', lambda x: x.median()),
+     ('Longest duration', lambda x: x.max()),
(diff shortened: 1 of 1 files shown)


twiecki:
Looks great. This actually made me wonder, is that a potential fix for that splits issue? That would filter out positions that have been open for a max period?

gusgordon:
I don't think so - just deleting round trips won't fix the issue. The error happens when a position is added or removed from the portfolio with no transaction. Let's say TSLA is added via spinoff, and the algorithm detects it has that position then closes it out immediately. From pyfolio's perspective, it didn't see the spinoff, so it thinks that close is the opening of a new short position. That short position round trip will then be seen as closed when the algo happens to decide to go long in that position. The same thing happens to whatever position the spinoff came from, but in reverse.

twiecki:
We add close-out orders to all open positions when the backtest ends, maybe we shouldn't do that. It wouldn't help with the second issue though.

PR #374, opened 2017-04-20
Update perf_stats table, update holdings plots, other various fixes

gusgordon:
* Take out information ratio from the performance statistics table, and add variance at risk and daily turnover. Also, use complete name of statistic (not just of the function) and add some percentage signs:
screen shot 2017-04-20 at 3 39 51 pm

 * Show min and max holdings in legend of long+short holdings plot.
 * Also some minor refactorings and doc fixes.

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -746,7 +746,7 @@ def perf_stats(returns, factor_returns=None, positions=None,
         for stat_func in FACTOR_STAT_FUNCS:
             res = stat_func(returns, factor_returns)
             stats[STAT_FUNC_NAMES[stat_func.__name__]] = res
-    
+
     for stat, value in stats.iteritems():
         if stat in STAT_FUNCS_PCT:
             stats[stat] = str(np.round(value * 100, 1)) + '%'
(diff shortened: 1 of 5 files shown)


pyfolio/timeseries.py
```diff
@@ -678,20 +735,27 @@ def perf_stats(returns, factor_returns=None, positions=None):
...
     if factor_returns is not None:
         for stat_func in FACTOR_STAT_FUNCS:
-            stats[stat_func.__name__] = stat_func(returns,
-                                                  factor_returns)
+            res = stat_func(returns, factor_returns)
+            stats[STAT_FUNC_NAMES[stat_func.__name__]] = res
+
+    for stat, value in stats.iteritems():
+        if stat in STAT_FUNCS_PCT:
+            stats[stat] = str(np.round(value * 100, 1)) + '%'
```
twiecki: I think we should do this in `show_perf_stats()` and leave these values float. Otherwise, if someone wants to do further computation with the output df from `perf_stats()` (like subtracting) it will fail for odd reasons.

twiecki:
See comment but otherwise looks great!

gusgordon:
@twiecki That makes sense to me, thanks for the review.

PR #375, opened 2017-04-24
Cutting of oos returns missed a day.

twiecki:


gusgordon:
Looks good, so if `live_start_date` was equal to a timestamp in the index then that was not included. We definitely want `returns_oos = returns[returns.index >= live_start_date]`, not `returns_is = returns[returns.index <= live_start_date]` instead (for the in-sample), right?

twiecki:
Correct, or there would be an overlap.

PR #382, opened 2017-05-15
Changes to accommodate for custom benchmarks in bayesian analysis.

bpajusco:
This will prevent MAP assignments to fall over due to NaNs in the benchmark series


twiecki:
Something must have gone wrong, it should show changes of the existing file, not add a new one.

PR #383, opened 2017-05-15
Changes to accommodate for custom benchmarks in bayesian analysis

bpajusco:
This will prevent MAP assignments to fall over due to NaNs in the benchmark series


twiecki:
You can probably fix this with `git rm bayesian.py`.

bpajusco:
Not idea why it fails Travis CI build. The file display changes I made on my local copies to handle the problem

twiecki:
`pyfolio/bayesian.py:96:54: E231 missing whitespace after ','`

Note that we enforce pep8 in the code-base, you can run flake8 to test.

Not sure this is the right fix, however. Can you remove the find_MAP() call and just try to call `pm.sample()` (without anything else) with the current master of pymc3?

bpajusco:
I've fixed the error you pointed out. As I said, I'm not a good coder at all and I'm only prototyping things but not ready for prime time at all.

As for your point about using the current code on master and just remove the MAP step when before sampling, the answer is that the whole model_return_t_alpha_beta() function simply won't work at all without my hack. If I didn't handle NaNs coming from index discrepancies in the returns series versus the chosen benchmark, the whole model would simply fall over even in calling np.linalg.lstsq(X, y)[0]. This is because X would include NaNs as the index in data would not be the same as the one in bmark. This isn't a problem with Quantopian supplied benchmark data (which if I'm not mistaken is coming from yahoo via pandas_datareader) but if I supplied my own index or had a series with a different holiday schedule than that of the benchmark I would end up with NaNs. The problem is evident when calling pm.sample() (with or without MAP) assignment but the reason why I mentioned MAP is because unless I fixed the data going into the optimiser I would get weird results when calling MAP. In other words, this isn't about MAP but a data issue that has repercussions throughout the code. I totally get the fact that there might be better ways to fix this but this is what I came up with and I haven't noticed weird alpha or beta values from running my version of the code.

twiecki:
Can't you call `.ffill()` before passing your data in?

twiecki:
I suppose the problem is that you have data for which no benchmark data exists. Do you have weekends in there or an odd start-time?

I think a better fix would be to do the intersection of benchmark and algo returns' index and use that.

bpajusco:
Most of the problem has to do with US holidays not being in the FF and benchmark factors used by pyfolio but in my return series. Your suggestion makes sense but I thought that forward filling (and back fillling in case the first day is a NaN too) was a way of not losing those days when we have a missing benchmark data since I care more about performance statistics of my series being correct than slightly messing up those of the benchmarks. 

At the end of the day, we should probably close the pull request since I realise how this solution is just a hack and I will let you know if I manage to figure out something better instead but I do think that I'm not the only one who has encountered a problem like this since not everybody only trades SPY and US based ETFs or similar. For example, I'm using futures traded in the US, Asia and Europe across multiple asset classes so I'm always going to have problems when matching against local benchmarks.

georgh0021:
@bpajusco closing PR as requested. Please let us know if you figure out a better solution!

@gusgordon @twiecki making pyfolio support multiple asset classes across multiple countries is too much to do at once, but we should make sure pyfolio supports US futures, since that sounds like the next big internal use case.

PR #389, opened 2017-05-25
Simple Tearsheet

georgh0021:
Added functions to create a simple tearsheet with only essential plots. Useful for beginners and people who just want a quick overview/basic visualization of their backtest.

Example simple tearsheet attached (using nonsense data).

simple_tearsheet_1
simple_tearsheet_2
simple_tearsheet_3
simple_tearsheet_4


twiecki:
Very cool! We talked about this before so great to see it realized here. Travis is broken due to pandas-datareader, so that's unrelated. Can you add a simple test to https://github.com/quantopian/pyfolio/blob/master/pyfolio/tests/test_tears.py?

georgh0021:
I added tests for all simple_tearsheet functions, but since Travis is broken I'm not sure if its working. I also touched up the code so it follows flake8 and takes in fewer parameters than the full_tearsheet functions, so its much easier to use.

@twiecki @gusgordon @Jstauth @adeychman thoughts and feedback?

gusgordon:
Looks great to me. Here are my suggestions:

- This tear sheet doesn't accept a `live_start_date` yet - in order for the cones to be plotted, that argument has to be passed to `plot_rolling_returns`.
- Probably no need for the flat returns plot (second one), the gross leverage plot, or the daily trading volume in shares plot. Those have overlapping information with other plots here. The top 10 holdings plot is also not super useful (we have discussed updating that in pyfolio before)
- I think the long/short holdings plot is slightly more useful than the `plot_holdings` plot, as it helps people further confirm their strategy is market neutral.
- As for things to add, maybe the transaction times distribution? I'm trying to think if there's anything else. Definitely open to suggestions on that, and my other points too.

Another thought is that we could have only one `simple_tear_sheet` function, instead of having the returns/positions/transactions components separate.  This would make it so your tear sheet was just one seamless image, and people could easily save it off, share it, etc. Any thoughts on that?

Jstauth:
on Gus's suggestions here are my thoughts:

- Probably no need for the flat returns plot (second one), the gross leverage plot, or the daily trading volume in shares plot. Those have overlapping information with other plots here. 
**Agree.**

- The top 10 holdings plot is also not super useful (we have discussed updating that in pyfolio before) I think the long/short holdings plot is slightly more useful than the plot_holdings plot, as it helps people further confirm their strategy is market neutral.

**I agree that plot as is isn't wildly useful, however as this is its the only plot that give you a sense of the position concentration and top N names, which is often a pretty useful first order thing to see about an algo. So until/unless we could replace it with something that gets to that info I'd be inclined to leave it in.**

As for things to add, maybe the transaction times distribution? 
**Agree**

georgh0021:
Took on board Jess's comments; the top 10 holdings plot is back in (along with a list of top 10 names).

Also, I've integrated all `create_simple_X_tear_sheet` into one `create_simple_tear_sheet`, so that the plot is one seamless jpg image.

Unfortunately, the table in the beginning isn't so easily put into that same jpg. 

Example attached.
screen shot 2017-05-26 at 2 39 28 pm
download-1

georgh0021:
Final edits done; waiting on green light from @twiecki to merge.

twiecki:
This looks great! I think we should add this to the tutorial as well. Either as a separate one, or under http://quantopian.github.io/pyfolio/notebooks/zipline_algo_example/.

gusgordon:
@twiecki The plan is to add the tutorial when George adds his factor tear sheet (which he's working on next), so merging this now.

PR #390, opened 2017-05-27
fixes #278

### LINKED ISSUE #278: Add rolling annual volatility plot
Similar to the rolling Sharpe Ratio plot, but just rolling annual volatility (which is simply the denominator from Sharpe Ratio).  As well, overlay the rolling annual volatility of SPY, say, in a lighter color, like gray.  This will allow easy visual comparison of an algo's volatility with the volatility of the US Stock Market (SPY)

hkopp:
Hi,
I have implemented running volatility and put it inside the returns_tear_sheet.
It does not compare the volatility with the volatility of SPY or something similar, as was originally requested in #278 

Oh, and for the three commits on Mar 07, I also do not know why they appear here, since they are already merged. Maybe I am using git in a wrong way.

Best,
hkopp


twiecki:
Thanks for the contribution!

Why not add the SPY volatility?

Can you paste a plot of that this looks like?

hkopp:
Actually I thought it would be really difficult. But you are right. I will take a shot.

hkopp:
Got it.
[Vol against SPY](https://cloud.githubusercontent.com/assets/6017673/26643508/d8e22488-4631-11e7-9241-221e8f3f580e.png)

pyfolio/plotting.py
```diff
@@ -896,7 +896,8 @@ def plot_rolling_beta(returns, factor_returns, legend_loc='best',
     return ax
 
 
-def plot_rolling_volatility(returns, rolling_window=APPROX_BDAYS_PER_MONTH * 6,
+def plot_rolling_volatility(returns, factor_returns=pd.DataFrame(),
```
twiecki: We usually just set it to None and check that below.

pyfolio/tears.py
```diff
@@ -361,7 +361,7 @@ def create_returns_tear_sheet(returns, positions=None,
         returns, benchmark_rets, ax=ax_rolling_beta)
 
     plotting.plot_rolling_volatility(
-        returns, ax=ax_rolling_volatility)
+        returns, benchmark_rets, ax=ax_rolling_volatility)
```
twiecki: best to use the kwarg explicitly.

pyfolio/plotting.py
```diff
@@ -896,6 +896,70 @@ def plot_rolling_beta(returns, factor_returns, legend_loc='best',
...
+        color='orangered',
+        linestyle='--',
+        lw=3)
+
+    ax.axhline(0.0, color='black', linestyle='-', lw=2)
+
+    ax.set_ylabel('Volatility')
+    ax.set_xlabel('')
+    if factor_returns.empty:
+        ax.legend(['Volatility', "Average Volatility"],
```
twiecki: Mixes ' and "

pyfolio/plotting.py
```diff
@@ -896,6 +896,70 @@ def plot_rolling_beta(returns, factor_returns, legend_loc='best',
...
+                  loc=legend_loc)
+    else:
+        ax.legend(['Volatility', 'Benchmark Volatility', "Average Volatility"],
```
twiecki: same.

pyfolio/timeseries.py
```diff
@@ -1024,6 +1024,28 @@ def gen_drawdown_table(returns, top=10):
...
+    rolling_vol_window : int
+        Length of rolling window, in days, over which to compute.
+
+    Returns
+    -------
+    pd.Series
+        Rolling volatility.
+    """
+
+    return returns.rolling(rolling_vol_window).std() \
```
twiecki: can't we use `annualized_volatility` directly here?

twiecki:
Left some comments, looks great otherwise.

pyfolio/plotting.py
hkopp: I compared with plot_rolling_beta one definition above, and there, the factor_returns are always required (despite the docstring saying that it is optional).
But yeah, I can fix this.

pyfolio/timeseries.py
hkopp: You mean annual_volatility()? This is deprecated. However, if you really insist I can do it.

pyfolio/plotting.py
hkopp: fixed

pyfolio/plotting.py
hkopp: fixed

pyfolio/tears.py
hkopp: fixed

pyfolio/plotting.py
twiecki: Ah, need to fix the doc-string above. And here would be great if you were to fix that.

pyfolio/timeseries.py
twiecki: The one from `empyrical` I mean.

hkopp: I think the one from empyrical does not work, since it returns a float and we need a pd.Series instead.

twiecki: Isn't that what returns.rolling requires?

hkopp: My rolling_volatility takes in the returns and the window as an integer. It returns a pd.Series.

Empyrical.annual_volatility also takes in the returns, but no window over which to compute the vol, since it does not return a pd.Series, but instead a float. Empyricals annual vol takes in a period as a string over which period to annualize, e.g., daily, weekly, ...

The only way I could use empyrical.annual_volatility in my function is, if I split the returns into sublists of length "rolling_vol_window" and compute empyrical.annual_volatility over each of those, which makes it really ugly and maybe also slow.

If you have a nicer idea on how to use empyrical, please tell me. You are more familiar with that, than I.

twiecki:
Thanks @hkopp. Seems like we have to resolve some pandas issues first.

twiecki:
Forked and extended in https://github.com/quantopian/pyfolio/pull/395. Thanks @hkopp.

georgh0021:
Kicking off a new build now that the pandas data reader issue is fixed.

georgh0021:
@hkopp all tests passed, as expected!

@twiecki I'm not sure if we want a test for the new `rolling_volatility` function: it's very simple, and nothing calls/depends on it (...yet). If so, we should [have one written on `test_timeseries.py`](https://github.com/quantopian/pyfolio/blob/master/pyfolio/tests/test_timeseries.py) before merging.

twiecki:
Thanks @hkopp and @georgh0021!

PR #391, opened 2017-06-12
Risk tear sheet

georgh0021:
Risk tear sheet: currently plots style factor weighted exposures, volume exposures, sector exposures.

To add: P&L attribution.


georgh0021:
Basic risk plots added, with docstrings. Unit tests performed. Attached are screenshots illustrating an example of a risk tear sheet.

To add: P&L attribution

screen shot 2017-06-13 at 9 42 09 am
screen shot 2017-06-13 at 9 42 20 am
screen shot 2017-06-13 at 9 42 41 am
screen shot 2017-06-13 at 9 42 30 am

pyfolio/tears.py
```diff
@@ -1217,3 +1229,161 @@ def create_bayesian_tear_sheet(returns, benchmark_rets=None,
...
+        - DataFrame with dates as index and equities as columns
+        - Example:
+                     Equity(24   Equity(62
+                       [AAPL])      [ABT])
+        2017-04-03	     311.0       206.0
+        2017-04-04	     311.0       206.0
+        2017-04-05	     311.0	     206.0
+
+    caps : pd.DataFrame
+        Daily Morningstar sector code per asset
```
gusgordon: Comment needs fixing

pyfolio/tears.py
```diff
@@ -1217,3 +1229,161 @@ def create_bayesian_tear_sheet(returns, benchmark_rets=None,
...
+        - Each item has dates as index, equities as columns
+        - Example item:
+                     Equity(24   Equity(62
+                       [AAPL])      [ABT])
+        2017-04-03	  -0.51284     1.39173
+        2017-04-04	  -0.73381     0.98149
+        2017-04-05	  -0.90132	   1.13981
+
+    sector : pd.DataFrame
+        Daily Morningstar sector code per asset
```
gusgordon: What if someone passes non-morningstar data? Maybe there's a way where they can pass strings instead of numbers, then we can plot the strings if so? Could also make that the default, then convert numbers to strings on our end before calling the tear sheet.

pyfolio/risk.py
```diff
@@ -0,0 +1,461 @@
...
+    grossed_frac = shares_grossed.divide(volumes)
+
+    longed_threshold = 100*longed_frac.quantile(percentile, axis=1)
+    shorted_threshold = 100*shorted_frac.quantile(percentile, axis=1)
+    grossed_threshold = 100*grossed_frac.quantile(percentile, axis=1)
+
+    return longed_threshold, shorted_threshold, grossed_threshold
+
+
+def plot_volume_exposures_longshort(longed_threshold, shorted_threshold,
```
gusgordon: Should this belong in the risk tear sheet, and not somewhere else? I'm also curious about why the number of shares transacted is useful to us, since typically, dollars traded is much more useful.

pyfolio/risk.py
```diff
@@ -0,0 +1,461 @@
...
+                 .format(100*percentile))
+    ax.set_xlabel('Date')
+    ax.set_ylabel('{}th Percentile of Proportion of Volume (%)'
+                  .format(100*percentile))
+    ax.legend()
+
+    return ax
+
+
+def plot_volume_exposures_gross(grossed_threshold, percentile, ax=None):
```
gusgordon: Same for this - I might have missed discussion where you guys talked about volume exposures?

pyfolio/risk.py
```diff
@@ -0,0 +1,461 @@
...
+        Name of style factor, for use in graph title
+    '''
+
+    if ax is None:
+        ax = plt.gca()
+
+    ax.plot(tot_sfe.index, tot_sfe, label=factor_name)
+    avg = tot_sfe.mean()
+    ax.axhline(avg, linestyle='-.', label='Mean = {:.3}'.format(avg))
+    ax.axhline(0, color='k', linestyle='-')
```
gusgordon: I'm still not convinced the best way to plot these is zoomed in with the 0 line bolded. Although the zoom is definitely nice, we aren't typically concerned with micro-movements of the signal and are more interesting in swings above of below zero. I think having a more standardized axis will be less confusing, and allow us to go through things more quickly. Do Jonathan or Jess prefer it the way you have it?

pyfolio/tears.py
```diff
@@ -1217,3 +1229,161 @@ def create_bayesian_tear_sheet(returns, benchmark_rets=None,
...
+                            [AAPL])           [ABT])
+        2017-04-03      34940859.00       4665573.80
+        2017-04-04	    35603329.10       4818463.90
+        2017-04-05	    41846731.75	      4129153.10
+
+    percentile : float
+        Percentile to use when computing and plotting volume exposures.
+        - Defaults to 10th percentile
+    '''
+
```
gusgordon: You should also pass `positions` through `check_intraday`, like here: https://github.com/quantopian/pyfolio/blob/master/pyfolio/tears.py#L603

Otherwise, for strategies that have no end-of-day holdings, your tear sheet will probably not be able to plot much.

That also requires passing transactions and returns to `create_risk_tear_sheet` :)

pyfolio/tears.py
georgh0021: Fixed!

pyfolio/risk.py
georgh0021: Specified by Jonathan: we take the number of shares held at the end of the day, divided by the total volume. That gives us a number per stock per day. We find the 10th percentile (i.e. the most illiquid) and plot that as a time series. So it is a measure of how exposed our portfolio is to illiquid stocks.

pyfolio/risk.py
georgh0021: I can center the zero line, but I don't think sharing y axes will be useful: the scales can be off by an order of magnitude, maybe even more.

pyfolio/risk.py
georgh0021: See comment above

georgh0021:
Several other comments:

- Make legends slightly opaque in the stackplots, so we can see the colors associated with the sectors/market caps
- Modify `create_risk_tear_sheet` so that the only required input is the positions: if nothing else is passed, simply plot nothing.

pyfolio/tears.py
georgh0021: Made functions so that users can optionally pass in a `sector_dict` that lists custom sector codes (keys, eg ints or strings) to sector names (values, must be strings). If no `sector_dict` is passed in, it defaults to Morningstar sector mappings.

pyfolio/tears.py
georgh0021: Done! Also did this for `create_simple_tear_sheet` too...

georgh0021:
Made edits from Gus' feedback; removed 'Date' labels x axis labels, and deleted `perf_attrib.py`: the PR is now ready to merge with master so that the risk tearsheet is up and running. The plan is to add performance attribution and P&L attribution at a later date.

gusgordon:
Kicked off another build now that #395 is merged.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import numpy as np
+import matplotlib.pyplot as plt
+from collections import OrderedDict
+
+SECTORS = OrderedDict([
+    (101, '101 Basic Materials'),
```
twiecki: Do we need the number in the string? Seems to just add clutter.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    (205, '205 Consumer Defensive'),
+    (206, '206 Healthcare'),
+    (207, '207 Utilities'),
+    (308, '308 Communication Services'),
+    (309, '309 Energy'),
+    (310, '310 Industrials'),
+    (311, '311 Technology')
+])
+
+CAP_CUTOFFS = [50000000, 300000000, 2000000000, 10000000000, 200000000000]
```
twiecki: This should also be a dict / OrderedDict / pd.Series.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+        - DataFrame with dates as index and equities as columns
+        - Example:
+                     Equity(24   Equity(62
+                       [AAPL])      [ABT])
+        2017-04-03	  -0.51284     1.39173
+        2017-04-04	  -0.73381     0.98149
+        2017-04-05	  -0.90132	   1.13981
+    '''
+
+    positions_wo_cash = positions.drop('cash', axis=1)
```
twiecki: I prefer `axis='columns'` for readability.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    gross_exposure = positions_wo_cash.abs().sum(axis=1)
+
+    sfe = positions_wo_cash.multiply(risk_factor, level=1) \
```
twiecki: `sfe` -> `style_factor_exposure` better to be verbose.

twiecki: Why is `level=1` required here? Shouldn't these just be single index?

twiecki: If these are mult-index, the doc string should reflect it. Moreover, multi-indexes should be named to give `level='factor'`.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    return tot_sfe
+
+
+def plot_style_factor_exposures(tot_sfe, factor_name, ax=None):
+    '''
+    Plots DataFrame output of compute_style_factor_exposures as a line graph
+
+    Parameters
+    ----------
+    tot_sfe : pd.Series
```
twiecki: verbose var names.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+        Daily style factor exposures (output of compute_style_factor_exposures)
+        - Time series with decimal style factor exposures
+        - Example:
+            2017-04-24    0.037820
+            2017-04-25    0.016413
+            2017-04-26   -0.021472
+            2017-04-27   -0.024859
+
+    factor_name : string
```
twiecki: That name should be in `pd.Series.name`.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    if ax is None:
+        ax = plt.gca()
+
+    ax.plot(tot_sfe.index, tot_sfe, label=factor_name)
+    avg = tot_sfe.mean()
+    ax.axhline(avg, linestyle='-.', label='Mean = {:.3}'.format(avg))
+    ax.axhline(0, color='k', linestyle='-')
+    _, _, y1, y2 = plt.axis()
+    a = max(abs(y1), abs(y2))
+    ax.set_ylim(-a, a)
```
twiecki: `ax.set(ylim=..., title=...)`

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    ax.set_title('Exposure to {}'.format(factor_name))
+    ax.set_ylabel('{} \n Weighted Exposure'.format(factor_name))
+    ax.legend()
+
+    return ax
+
+
+def compute_sector_exposures(positions, sectors, sector_dict=None):
```
twiecki: `sector_dict=SECTORS` (SECTORS should be a frozen dict).

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    gross_exposures = []
+    net_exposures = []
+
+    positions_wo_cash = positions.drop('cash', axis=1)
+    long_exposure = positions_wo_cash[positions_wo_cash > 0].sum(axis=1)
+    short_exposure = positions_wo_cash[positions_wo_cash < 0].abs().sum(axis=1)
+    gross_exposure = positions_wo_cash.abs().sum(axis=1)
+    sectors.columns = sectors.columns.astype(int)
+
+    for sector_id in sector_ids:
```
twiecki: If you convert positions and sectors to long, concatenate them to give:
```
             position, sector
dt, symbol
2011-1-1, 'AAPL', 100, 'Energy'
...
```

The below should just be a single `.groupby()` call.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    long_exposures = []
+    short_exposures = []
+    gross_exposures = []
+    net_exposures = []
+
+    positions_wo_cash = positions.drop('cash', axis=1)
+    long_exposure = positions_wo_cash[positions_wo_cash > 0].sum(axis=1)
+    short_exposure = positions_wo_cash[positions_wo_cash < 0].abs().sum(axis=1)
+    gross_exposure = positions_wo_cash.abs().sum(axis=1)
+    sectors.columns = sectors.columns.astype(int)
```
twiecki: Instead I would drop the int mapping as soon as possible. There really is no reason to use codes when we could just use the names directly. I would go even further and suggest dropping the `sector_map` kwarg and expecting the input to have the right names already, or at least make the map optional.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    '''
+    Plots outputs of compute_sector_exposures as area charts
+
+    Parameters
+    ----------
+    long_exposures, short_exposures : arrays
+        Arrays of long and short sector exposures (output of
+        compute_sector_exposures).
+
+    sector_dict : dict or OrderedDict
```
twiecki: See above.

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+
+    if ax is None:
+        ax = plt.gca()
+
+    if sector_dict is None:
+        sector_names = SECTORS.values()
+    else:
+        sector_names = sector_dict.values()
+
+    colors = ['#FF9999', '#FFCC99', '#FFFF99', '#CCFF99', '#99FF99', '#99FFCC',
```
twiecki: Isn't there a colormap we can use instead?

pyfolio/risk.py
```diff
@@ -0,0 +1,486 @@
...
+    ax.stackplot(gross_exposures[0].index, gross_exposures,
+                 labels=CAP_NAMES, colors=colors, baseline='zero')
+    ax.axhline(0, color='k', linestyle='-')
+    ax.set_title('Gross Exposure to Market Caps')
+    ax.set_ylabel('Proportion of Gross Exposure \n in Market Cap Buckets')
+
+    return ax
+
+
+def plot_cap_exposures_net(net_exposures, ax=None):
```
twiecki: Why do we need separate plotting functions for all different types of exposures? Couldn't a single function just plot whatever is passed in?

georgh0021: @twiecki no, different types of exposures are plotted differently. E.g. long/short exposures are plotted at stack plots that go positive and negative, gross exposures are plotted as stack plots that are only positive (and so half the size) and net exposures are plotted as line graphs. It's possible to have a single function that plots everything, and figures out what you passed in, but have a separate function for each seemed cleaner.

pyfolio/risk.py
georgh0021: I might be wrong about this, but [all of the qualitative colormaps](https://matplotlib.org/examples/color/colormaps_reference.html) with 11+ colors (there are 11 sectors) had visually indistinct colors (e.g. different hues of the blue), which I found undesirable

joshpayne:
Hi, I'm going to stick my nose into this data viz work :)

For the "rainbow" charts for sector exposures I find it nearly impossible to read the chart and understand which sector is which. The legend is long and difficult to use.

Can you:
1) actually label the color segments themselves and remove the legend? so the graph 'chunk' itself would be labeled with it's sector.

If that's not possible given the tools we have here
2) can you reverse sort the legend so that the order of the legend matches the order of the chart. Right now it looks like they're totally opposite.

pyfolio/risk.py
georgh0021: Right as always, Thomas. `level=1` doesn't do anything: it must be an artifact of some hacked-together code I wrote.

pyfolio/risk.py
georgh0021: Don't want to be difficult, but tidying `positions` and `sectors` to long format seems like a lot of work for essentially the same result that the current code gives. Is there a particular reason why converting to long format would be more advantageous?

twiecki: Maintainability. Mostly, this is part of a larger question of input data structures.

pyfolio/risk.py
georgh0021: It's actually `pd.DataFrame.name`... I wasn't aware that it was good practice to specify a value for the `DataFrame.name` attribute. I imagine that would get cumbersome when the number of style factors we wish to plot gets larger.

What I will do is make `factor_name` default to `tot_style_factor_exposure.name`, but allow the user to optionally override this by passing in their own style factor name.

georgh0021: Docstring also updated to reflect this change

pyfolio/risk.py
georgh0021: Did not know that `ax` had that function! Replaced all instances of `ax.set_foo(...); ax.set_bar(...)` with `ax.set(foo=..., bar=...)

georgh0021:
Finally done! Ready to merge @gusgordon

twiecki:
I'm not quite done yet.

georgh0021:
Sorry, didn't realize. Will hold off on merging until further feedback

pyfolio/risk.py
georgh0021: I'm all for using long data, but I want to point out that if our input data structures change, this isn't the only piece of code that will need refactoring: all `compute_foo()` functions will have maintenance issues.

pyfolio/risk.py
twiecki: What about Set3?

georgh0021: Didn't know that existed! Changed it so that `color_list = plt.cm.Set3(np.linspace(...))` for all multicolored graphs. Goodbye rainbow stack plots.

pyfolio/tears.py
```diff
@@ -1217,3 +1233,171 @@ def create_bayesian_tear_sheet(returns, benchmark_rets=None,
     plt.show()
     if return_fig:
         return fig
+
+
+@plotting_context
+def create_risk_tear_sheet(positions,
+                           style_factor_panel=None,
+                           sectors=None,
```
georgh0021: Bug: what if the input data do not have the same DateTimeIndexes? It is very likely that the risk metrics data will start before and end later than the positions data. If so, the tear sheet produces misleading graphs and artificially deflated means as below. I.e. the analysis returns 0 until all data inputs have non-zero/non-nan values.

One fix is to only analyze/plot over the time range where all data inputs are non-zero/non-nan. I.e. in `create_risk_tear_sheet`, mask all data inputs to the intersection of their DateTimeIndexes. @gusgordon @twiecki thoughts?

screen shot 2017-07-08 at 9 42 56 pm

pyfolio/risk.py
georgh0021: Actually, I prefer to have the colors occur in a rainbow: the reason being that if you were looking at a "patch" on the graph and wanted to look up sector in the legend, the order of the rainbow would help you quickly find the correct sector, whereas `Set3` orders colors in an essentially random order, forcing the reader to look through every legend entry and giving a greater cognitive load on the reader.

I'll use `gist_rainbow` colormap instead of `Set3`. It's quite a garish palette, so I'll set `alpha=0.8`.

twiecki:
Some more general thoughts:
* Before we had utility functions in e.g. `positions.py` and plotting functions in `plotting.py`. Here we are grouping the math and plotting functions into the same file. I feel like we should stay consistent. Our options are:
   1. Move all plotting functions into `plotting.py`
   2. Remove `plotting.py` and move things into the other files
* Data format: There main problem we need to solve is that this PR uses `pandas.Panel` which is deprecated (https://pandas.pydata.org/pandas-docs/stable/dsintro.html#panel). Thus, at a minimum we should change the Panels to multiindex dataframes. Once we have done this we could go one step further and convert all data to long (I realize this is a major refactor) where dt and sid are the index and columns are position, sector, cap, momentum, reversal, shares_held, energy etc. In that world we only have a single data structure to be passed internally, the risk tear sheet could check for the existence of certain columns and plotting functions extract a list of columns they want to plot. We have ultimately switched to the same format in `alphalens` under a much higher cost because we have done so much later in the game, but it definitely was worth it, as this format is also much easier to extend. I also think we will want to do something similar for the risk model output as well as for the PnL attribution, so the earlier we bite the bullet the better.

Thoughts @georgh0021, @gusgordon?

georgh0021:
As of right now we have three issues that need addressing:

1. Non-identical dt indexes produce misleading graphs (pretty major bug)
2. Moving code around into `risk.py` and `plotting.py` (for consistency)
3. Data format: a) deprecating panels and b) changing to long format (for extendability)

1 and 3a have straightforward solutions: those will be my tasks over the next few days.

Wrt 2, I agree we need to stay consistent, but I think it makes more sense to do away with `plotting.py` and move all plotting functions into their respective modules (reasons below). If we decide on this approach, nothing needs to be done for this PR, but we'll need to open a new one to break up `plotting.py`.

- pyfolio is modularized into separate tear sheets because that's how the user will consume the final product: tear sheet by tear sheet. It doesn't make sense to me to have the functionality of, say, the returns tear sheet split between two different modules.
- pyfolio plots much more than it computes: `plotting.py` clocks in at around 2000 lines of code, whereas most other modules barely make 500. Refactoring so that plotting functions immediately follow their corresponding compute functions would make pyfolio much easier to read.

Wrt 3b, the spirit of long data goes way beyond the risk tear sheet: another long term PR would be to refactor all pyfolio code (i.e. from the returns tear sheet up) so that it's long data in, long data out. This will probably have to wait a while.

@gusgordon?

gusgordon:
@twiecki @georgh0021 Agree on removing plotting.py and using the new data format. I vote we should merge this PR (once any loose ends are cleaned up), then we can open another PR where all of that refactoring is done. Do you guys think that makes sense?

twiecki:
Agreed on plotting.py.

I don't agree on 3b. I don't think other parts of pyfolio need to be converted to long, there is only returns (series), positions (wide), and transactions (long), these all seem fine. It's only affecting this PR.

georgh0021:
Makes sense. I'll see if I can fix 1) by the end of today, and then we can open up a separate PR to refactor. @twiecki does that sound good?

twiecki:
OK, that's fine.

georgh0021:
Fixed problem 1; ready to merge as soon as builds finish @gusgordon @twiecki

gusgordon:
@georgh0021 test failed again for the same reason. I would see if you can reproduce it on your computer. Make sure your branch is up to date and rebased with this one (you could just pull pyfolio into a new folder and checkout this branch).

georgh0021:
@twiecki there was a bug in test_timeseries.py caused by pandas data reader moving from Yahoo to Google: basically, there was only one drawdown in a test that tested for overlapping drawdowns. It should be fixed now.

PR #395, opened 2017-06-29
Fix #385 among other things

### LINKED ISSUE #385: Yahoo URL Error
When the following is run
```
%matplotlib inline
import pyfolio as pf
import pandas as pd
import numpy as np
stock_rets = pf.utils.get_symbol_rets('AGG')
```
I get a RemoteDataError for the baked-in URL.  Apparently Yahoo has made some recent changes to their API's
```
---------------------------------------------------------------------------
RemoteDataError                           Traceback (most recent call last)
<ipython-input-3-fd2a04977723> in <module>()
      3 import pandas as pd
      4 import numpy as np
----> 5 stock_rets = pf.utils.get_symbol_rets('AGG')

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pyfolio/utils.py in get_symbol_rets(symbol, start, end)
    473     return SETTINGS['returns_func'](symbol,
    474                                     start=start,
--> 475                                     end=end)
    476 
    477 

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pyfolio/utils.py in default_returns_func(symbol, start, end)
    259         rets = rets[start:end]
    260     else:
--> 261         rets = get_symbol_from_yahoo(symbol, start=start, end=end)
    262 
    263     return rets[symbol]

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pyfolio/utils.py in get_symbol_from_yahoo(symbol, start, end)
    212         Returns of symbol in requested period.
    213     """
--> 214     px = web.get_data_yahoo(symbol, start=start, end=end)
    215     rets = px[['Adj Close']].pct_change().dropna()
    216     rets.index = rets.index.tz_localize("UTC")

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/data.py in get_data_yahoo(*args, **kwargs)
     31 
     32 def get_data_yahoo(*args, **kwargs):
---> 33     return YahooDailyReader(*args, **kwargs).read()
     34 
     35 def get_data_yahoo_actions(*args, **kwargs):

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/yahoo/daily.py in read(self)
     75     def read(self):
     76         """ read one data from specified URL """
---> 77         df = super(YahooDailyReader, self).read()
     78         if self.ret_index:
     79             df['Ret_Index'] = _calc_return_index(df['Adj Close'])

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/base.py in read(self)
    171         # If a single symbol, (e.g., 'GOOG')
    172         if isinstance(self.symbols, (compat.string_types, int)):
--> 173             df = self._read_one_data(self.url, params=self._get_params(self.symbols))
    174         # Or multiple symbols, (e.g., ['GOOG', 'AAPL', 'MSFT'])
    175         elif isinstance(self.symbols, DataFrame):

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/base.py in _read_one_data(self, url, params)
     78         """ read one data from specified URL """
     79         if self._format == 'string':
---> 80             out = self._read_url_as_StringIO(url, params=params)
     81         elif self._format == 'json':
     82             out = self._get_response(url, params=params).json()

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/base.py in _read_url_as_StringIO(self, url, params)
     89         Open url (and retry)
     90         """
---> 91         response = self._get_response(url, params=params)
     92         out = StringIO()
     93         if isinstance(response.content, compat.binary_type):

/home/ubuntu/anaconda3/envs/idpf/lib/python3.5/site-packages/pandas_datareader/base.py in _get_response(self, url, params)
    115             time.sleep(self.pause)
    116 
--> 117         raise RemoteDataError('Unable to read URL: {0}'.format(url))
    118 
    119     def _read_lines(self, out):

RemoteDataError: Unable to read URL: http://ichart.finance.yahoo.com/table.csv
```

twiecki:
There were some problems after updating pandas and pymc3 3.1.

### DIFF
--- a/pyfolio/tests/test_pos.py
+++ b/pyfolio/tests/test_pos.py
@@ -11,8 +11,7 @@ from pandas import (
     Timestamp,
     read_csv
 )
-from pandas.util.testing import (assert_frame_equal,
-                                 assert_equal)
+from pandas.util.testing import assert_frame_equal
 
 from numpy import (
     arange,
@@ -166,7 +165,7 @@ class PositionsTestCase(TestCase):
     ])
     def test_detect_intraday(self, positions, transactions, expected):
         detected = detect_intraday(positions, transactions, threshold=0.25)
-        assert_equal(detected, expected)
+        assert detected == expected
 
     @parameterized.expand([
         ('infer', test_returns, test_pos, test_txn, test_pos),
@@ -184,4 +183,4 @@ class PositionsTestCase(TestCase):
     def test_estimate_intraday(self, returns, positions,
                                transactions, expected):
         intraday_pos = estimate_intraday(returns, positions, transactions)
-        assert_equal(intraday_pos.shape, expected)
+        assert intraday_pos.shape == expected
--- a/pyfolio/tests/test_tears.py
+++ b/pyfolio/tests/test_tears.py
@@ -26,7 +26,7 @@ class PositionsTestCase(TestCase):
     test_returns = read_csv(
         gzip.open(
             __location__ + '/test_data/test_returns.csv.gz'),
-        index_col=0, parse_dates=True).iloc[4:]
+        index_col=0, parse_dates=True)
     test_returns = to_series(to_utc(test_returns))
     test_txn = to_utc(read_csv(
         gzip.open(
--- a/pyfolio/tests/test_tears.py
+++ b/pyfolio/tests/test_tears.py
@@ -26,7 +26,7 @@ class PositionsTestCase(TestCase):
     test_returns = read_csv(
         gzip.open(
             __location__ + '/test_data/test_returns.csv.gz'),
-        index_col=0, parse_dates=True)
+        index_col=0, parse_dates=True).iloc[4:]
     test_returns = to_series(to_utc(test_returns))
     test_txn = to_utc(read_csv(
         gzip.open(
--- a/setup.py
+++ b/setup.py
@@ -53,11 +53,11 @@ install_reqs = [
     'scipy>=0.14.0',
     'seaborn>=0.7.1',
     'pandas-datareader>=0.2',
-    'empyrical>=0.2.2'
+    'empyrical>=0.3.0'
 ]
 
 test_reqs = ['nose>=1.3.7', 'nose-parameterized>=0.5.0', 'runipy>=0.1.3']
-bayesian_reqs = ['pymc3']
+bayesian_reqs = ['pymc3 >= 3.1']
 
 extras_reqs = {
     'bayesian': bayesian_reqs,
--- a/pyfolio/bayesian.py
+++ b/pyfolio/bayesian.py
@@ -18,7 +18,6 @@ import numpy as np
 import pandas as pd
 import scipy as sp
 from scipy import stats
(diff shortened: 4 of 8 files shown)


pyfolio/tests/test_pos.py
```diff
@@ -166,7 +165,7 @@ def test_max_median_exposure(self, positions, expected):
     ])
     def test_detect_intraday(self, positions, transactions, expected):
         detected = detect_intraday(positions, transactions, threshold=0.25)
-        assert_equal(detected, expected)
+        assert detected == expected
```
gusgordon: Why do we use `assert` here an not one of the pandas assertion functions?

setup.py
```diff
@@ -53,11 +53,11 @@
     'scipy>=0.14.0',
     'seaborn>=0.7.1',
     'pandas-datareader>=0.2',
-    'empyrical>=0.2.2'
+    'empyrical>=0.3.0'
 ]
```
gusgordon: Do we also want to pin versions instead of using `>=`? The current pandas issues were due to this, but there are advantages/problems with both `>=` and `==`.

pyfolio/tests/test_tears.py
```diff
@@ -26,7 +26,7 @@ class PositionsTestCase(TestCase):
     test_returns = read_csv(
         gzip.open(
             __location__ + '/test_data/test_returns.csv.gz'),
-        index_col=0, parse_dates=True)
+        index_col=0, parse_dates=True).iloc[4:]
```
gusgordon: Why do we skip the first 4 rows for the returns CSV, but don't skip rows for the positions or transactions CSVs?

gusgordon:
Looks great, thank you @twiecki!

pyfolio/tests/test_pos.py
twiecki: Seems like assert_equal was removed.

setup.py
twiecki: Maybe you're right.

pyfolio/tests/test_tears.py
twiecki: those were all 0's. it was part of me tracking down a problem but I think the problem was somewhere else so we can try removing this again.

setup.py
richafrank: My instinct is to allow a range. Otherwise every environment/application that uses pyfolio must use exactly this version of empyrical. This means we couldn't install zipline with a newer version of empyrical on our research platform, if we wanted newer empyrical features. If we can trust the maintainers of empyrical to keep its API backwards compatible, then we should be ok 😉 .

I'm not sure what the pandas issues were, though, so I might be off base here.

gusgordon: That makes sense, thanks @richafrank

twiecki:
Should probably cut a new release. @richafrank anyone who can help with that?

PR #398, opened 2017-07-03
PnL Tear Sheet

georgh0021:
Basic PnL attribution tear sheet: breaks down PnL by market cap, sector, Q universe and entering price.

The ultimate goal is to incorporate complete algo performance attribution into this tear sheet.


georgh0021:
The PnL breakdown bar charts should be added to perf attrib tear sheet on #400

PR #400, opened 2017-07-11
performance attribution tear sheet

georgh0021:
Performance attribution gets the auspicious reference number 400!

To complement the newly-merged risk tear sheet, we would like pyfolio to be able to do performance attribution: in other words, to answer the question "how much of today's PnL came from mean reversion; how much was driven by the healthcare sector; how much of it was due to alpha?". In addition, we would like pyfolio to be amenable to analysis using factor risk models: both Q's in-house risk model, as well as other risk models developed by our users.

This is a work in progress: below is a core dump of what we would like the end result to look like. The plots below analyze an example momentum algorithm.

I'll edit this post as the tear sheet takes on a more and more solid form.

screen shot 2017-07-12 at 2 49 46 pm
screen shot 2017-07-12 at 2 50 12 pm
screen shot 2017-07-12 at 2 50 22 pm
screen shot 2017-07-12 at 2 50 40 pm
screen shot 2017-07-12 at 2 50 48 pm
screen shot 2017-07-12 at 2 50 56 pm


georgh0021:
As discussed in #391, we'd like to move to long-format data structures, for ease of extendibility. This is a major refactor, but it is better to bite the bullet sooner rather than later. I'll work on this now.

georgh0021:
Closing in favor of #417 and future branch PRs.

PR #406, opened 2017-07-31
Fama-French multivariate regression

### LINKED ISSUE #379: Fama-French should use multivariate regression
https://github.com/quantopian/pyfolio/blob/master/pyfolio/timeseries.py#L563 shows we are using separate linear regressions which is problematic because the factors are confounded. 

We should instead use multivariate regression: http://stackoverflow.com/questions/19991445/run-an-ols-regression-with-pandas-data-frame

CC @jlowin

eigenfoo:
Fixes #379 

We don't just want a multivariate regression, we want a _rolling_ multivariate regression. Pandas used to support this sort of thing with `pd.stats.ols.MovingOLS`, but that has unfortunately been deprecated.

A solution is described [on StackOverflow](https://stackoverflow.com/questions/37317727/deprecated-rolling-window-option-in-ols-from-pandas-to-statsmodels), but frustratingly this solution doesn't work since `apply` only works on Series data (see citynorman's comment on the top answer). So, we will have to write our own rolling multivariate regression...


georgh0021:
@twiecki the Python 3.4 build is very sad without the `statsmodels` module

```
  File "/home/travis/build/quantopian/pyfolio/pyfolio/timeseries.py", line 24, in <module>
    import statsmodels.formula.api as sm
ImportError: No module named 'statsmodels'
```

georgh0021:
@twiecki @gusgordon the Python 3.4 build seems very sad without the `statsmodels` module

```
  File "/home/travis/build/quantopian/pyfolio/pyfolio/timeseries.py", line 24, in <module>
    import statsmodels.formula.api as sm
ImportError: No module named 'statsmodels'
```

pyfolio/timeseries.py
```diff
@@ -580,8 +584,27 @@ def rolling_fama_french(returns, factor_returns=None,
...
+
+    # have NaNs when there is insufficient data to do a regression
+    regression_coeffs = np.empty((len(regression_df.index[:rolling_window]),
+                                  len(factor_returns.columns)))
+    regression_coeffs.fill(np.nan)
+
+    for beg, end in zip(regression_df.index[:-rolling_window],
+                        regression_df.index[rolling_window:]):
+        window = regression_df.loc[beg:end]
+        coeffs = sm.ols(formula='rets ~ SMB + HML + UMD - 1', data=window) \
```
georgh0021: I also want to make sure that this is indeed the Fama-French betas. Is this computation correct? Note that the `-
 1` in the formula means to set the intercept equal to 0.

pyfolio/timeseries.py
```diff
@@ -580,8 +584,27 @@ def rolling_fama_french(returns, factor_returns=None,
...
+            .fit().params.values
```
georgh0021: I also want to make sure that this computation is correct. Are these parameters the Fama French betas?

The `-1` in the `formula` keyword means to set the intercept equal to 0.

twiecki: we should not force the intercept to be 0.

twiecki:
Yes, we need to add statsmodels to the requirements file.

pyfolio/timeseries.py
georgh0021: Fixed!

georgh0021:
Added `statsmodels` to requirements file and fixed bug forcing regression intercept to be 0. The Travis builds seem to be passing, although the Python 3.4 build timed out. Kicking off a new Python 3.4 build now, and will check back again tomorrow morning.

One small concern: on a jupyter notebook, `coeffs` was a `pd.Series`. However, Travis raised several errors, and claimed that `coeffs` was actually a numpy `ndarray`. I'm just worried that this might be a versioning issue with the newly added `statsmodels` dependency...

Other than that, the PR is ready for review and merge! @gusgordon @twiecki

georgh0021:
@twiecki the Travis builds are not failing, but are timing out. [See here](https://travis-ci.org/quantopian/pyfolio/builds/259619073?utm_source=github_status&utm_medium=notification) for an example. I've tried implementing a solution using `travis_wait`, but that does not seem to help: builds still time out. @gusgordon and I don't really know what's going on. Any thoughts?

twiecki:
@georgh0021 how slow is the regression when you try it locally?

georgh0021:
```
In [13]: %timeit rolling_fama_french(returns, factor_returns)
         1 loop, best of 3: 4.52 s per loop
```

twiecki:
4.52 secs is quite slow for this simple functionality. How long a time-range did you check this on? The numpy version is not ideal but perhaps our best shot if we want to keep the functionality.

twiecki:
Why do you test on only one factor? `factor_returns.iloc[:, 0],`

georgh0021:
Laziness 😛  the code on StackOverflow only took in one variable, I just copied and pasted to see what it would look like. Will refactor and test today.

twiecki:
Not sure I trust that code. Have you tried `sklearn.linear_model.LinearRegression`?

georgh0021:
@twiecki `sklearn` runs much faster!
``` 1 loop, best of 3: 479 ms per loop ```

Travis can't seem to find the package though.
```
PackageNotFoundError: Packages missing in current channels:
            
  - sklearn
We have searched for the packages in the following channels:
            
  - https://repo.continuum.io/pkgs/free/linux-64
  - https://repo.continuum.io/pkgs/free/noarch
  - https://repo.continuum.io/pkgs/r/linux-64
  - https://repo.continuum.io/pkgs/r/noarch
  - https://repo.continuum.io/pkgs/pro/linux-64
  - https://repo.continuum.io/pkgs/pro/noarch
            
The command "conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy flake8 seaborn sklearn runipy pytables networkx pandas-datareader matplotlib-tests joblib" failed and exited with 1 during .
Your build has been stopped.
```

twiecki:
Great! it's `conda install scikit-learn`

georgh0021:
@twiecki @gusgordon Travis builds are still timing out 😢

georgh0021:
@richafrank @twiecki bump. No idea why the tests are timing out.

twiecki:
Hm, seems like 3.4 and 3.5 are pretty close to the limit. At this point we can probably drop 3.4 all-together.

eigenfoo:
Not sure why the builds aren't timing out now... Something to keep in mind going forward I suppose. Once this becomes a serious problem we can look into finding a solution.

twiecki:
Well, I think it's right on the razor's edge. We should probably just test over a shorter time-period.

PR #408, opened 2017-08-01
Compute absolute capital in gross leverage denominator

twiecki:
Current implementation implements net exposure.


georgh0021:
Looks like quite a few tests will need to be rewritten

twiecki:
Actually it was correct before because shorts end up as cash.

PR #411, opened 2017-08-09
Fix example notebooks

### LINKED ISSUE #410: Example notebooks don't work
Some of the example notebooks are saved with the correct graphical output, but throw an error when you actually try to run the cells. We should definitely clean this up to make it easier for people to learn how to use pyfolio.

eigenfoo:
Fixes #410


twiecki:
I don't think this NB should be here in the first place and we should delete it.

georgh0021:
I'm also cleaning up the other notebooks: some are very old, and are throwing errors because pyfolio has changed so much. Will delete the linear algebra tests notebook, though.

georgh0021:
Two issues: the pyfolio talk slides have the same picture for a "traditional hedge fund" and "crowd-sourced hedge fund", so we embarrassingly just liken ourselves to a normal hedge fund. Contacted marketing for the slides.

Also, the zipline example notebook is now broken due to changes like the yahoo data reader. Contacted academia for someone with more experience with zipline to rewrite the notebook.

georgh0021:
Fixed the hedge fund pngs in the pyfolio talk slide. Just waiting on the zipline example notebook from @nathanwolfe now.

georgh0021:
Notebooks all cleaned up

@twiecki @richafrank requesting review and sign off (no changes to API or dependencies)

twiecki:
Looks great. You also need to fix the links in https://github.com/quantopian/pyfolio/tree/master/docs/notebooks.

richafrank:
Looks safe from my end. 👍

gusgordon:
This is great!

For the zipline example maybe we can use the newer IPython magic: https://github.com/quantopian/zipline/blob/master/zipline/examples/buyapple.ipynb

Also, trying to think if there's any way we can de-clutter this further. If people are first looking at pyfolio and want an example, it's not going to be obvious to them which notebook gives the best general overview. I wonder if there's a way we can combine a few of the notebooks or something.

georgh0021:
I believe we are using the `%load_ext` magic, but I might not be the best person to talk to about this... @nathanwolfe might be able to help here.

About decluttering: the notebooks don't have very many cells, it's just the tear sheets themselves that are taking up the space... I actually think that combining notebooks would make it harder to figure out what's going on. Perhaps we should just name the notebooks more descriptively?

eigenfoo:
This PR is ready to merge btw

twiecki:
What about https://github.com/quantopian/pyfolio/pull/411#issuecomment-322171642?

eigenfoo:
Completely forgot! I made a new PR for that

PR #417, opened 2017-08-16
ENH: add perf attribution + summary table

vikram-narayan:
add basic perf attribution and print a summary table

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -36,12 +36,17 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings):
     positions: pd.DataFrame
         Daily holdings (in dollars or percentages), indexed by date.
         Will be converted to percentages if positions are in dollars.
-        - Example:
-                            AAPL       TLT       XOM
-            2017-01-01  0.408046  0.534483  0.057471
-            2017-01-02  0.449367  0.101266  0.449367
-            2017-01-03  0.302632  0.282895  0.414474
+        Short positions show up as cash in the 'cash' column.
+        - Examples:
+                        AAPL  TLT  XOM  cash
+            2017-01-01    34   58   10     0
+            2017-01-02    22   77   18     0
+            2017-01-03   -15   27   30    15
 
+                            AAPL       TLT       XOM  cash
+            2017-01-01  0.333333  0.568627  0.098039   0.0
+            2017-01-02  0.188034  0.658120  0.153846   0.0
+            2017-01-03  0.208333  0.375000  0.416667   0.0
 
     factor_returns : pd.DataFrame
         Returns by factor, with date as index and factors as columns
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -37,10 +37,11 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings):
         Daily holdings (in dollars or percentages), indexed by date.
         Will be converted to percentages if positions are in dollars.
         - Example:
-                        AAPL  TLT  XOM
-            2017-01-01    71   93   10
-            2017-01-02    71   16   71
-            2017-01-03    46   43   63
+                            AAPL       TLT       XOM
+            2017-01-01  0.408046  0.534483  0.057471
+            2017-01-02  0.449367  0.101266  0.449367
+            2017-01-03  0.302632  0.282895  0.414474
+
 
     factor_returns : pd.DataFrame
         Returns by factor, with date as index and factors as columns
@@ -77,7 +78,6 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings):
     positions = get_percent_alloc(positions)
     positions = positions.stack()
     positions.index = positions.index.set_names(['dt', 'ticker'])
-
     risk_exposures = factor_loadings.multiply(positions,
                                               axis='rows')
     risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
@@ -90,7 +90,7 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings):
                                'common_returns': common_returns,
                                'specific_returns': specific_returns})
 
-    return  pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
+    return pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
 
 
 def create_perf_attrib_stats(perf_attrib):
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -1,0 +1,149 @@
+#
+# Copyright 2017 Quantopian, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import division
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+
+from pyfolio.utils import print_table
+
+
+def perf_attrib(factor_loadings,
+                factor_returns,
+                strategy_daily_returns,
+                strategy_daily_holdings):
+    """
+    Does performance attribution given risk info.
+
+    Parameters
+    ----------
+    factor_loadings : pd.DataFrame
+        Factor loadings for all days in the date range, with date and ticker as
+        index, and factors as columns.
+        - Example:
+                               momentum  reversal
+            dt         ticker
+            2017-01-01 AAPL   -1.592914  0.852830
+                       TLT     0.184864  0.895534
+                       XOM     0.993160  1.149353
+            2017-01-02 AAPL   -0.140009 -0.524952
+                       TLT    -1.066978  0.185435
+                       XOM    -1.798401  0.761549
+
+    factor_returns : pd.DataFrame
+        Returns by factor, with date as index and factors as columns
+        - Example:
+                        momentum  reversal
+            2017-01-01  0.002779 -0.005453
+            2017-01-02  0.001096  0.010290
+
+    strategy_daily_returns : pd.DataFrame
+        Returns for each day in the date range.
+        - Example:
+                            AAPL       TLT       XOM
+            2017-01-01 -0.014092 -0.003938 -0.001508
+            2017-01-02  0.011921  0.003039 -0.001546
+
+    strategy_daily_holdings: pd.Series
+        Daily holdings for all days in the date range, indexed by date
+        and ticker
+        - Example:
+            dt          ticker
+            2017-01-01  AAPL      71
+                        TLT       93
+                        XOM       10
+            2017-01-02  AAPL      71
+                        TLT       16
+                        XOM       71
+
+    Returns
+    -------
+    exposures : pd.DataFrame
+        df with factors as columns, and datetimes as index
+        - Example:
+                                 momentum    reversal
+            dt         ticker
+            2017-01-01 AAPL   -113.096901   60.550926
+                       TLT      17.192383   83.284658
+                       XOM       9.931600   11.493529
+            2017-01-02 AAPL     -9.940674  -37.271562
+                       TLT     -17.071643    2.966952
+                       XOM    -127.686503   54.070000
+
+    perf_attribution : pd.DataFrame
+        df with factors, common returns, and specific returns as columns,
+        and datetimes as index
+        - Example:
+                        momentum  reversal  common_returns  specific_returns
+            dt
+            2017-01-01  0.249087  0.935925        1.185012          1.185012
+            2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
+    """
+
+    risk_exposures = factor_loadings.multiply(strategy_daily_holdings,
+                                              axis='rows')
+    risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
+    perf_attrib_style = risk_exposures_portfolio.multiply(factor_returns)
+
+    common_pa = perf_attrib_style.sum(axis=1)
+    total_returns_daily = strategy_daily_returns.sum(axis=1)
+    specific_pa = total_returns_daily - common_pa
+
+    perf_attrib_common = perf_attrib_style.assign(common_returns=common_pa)
+    perf_attrib = perf_attrib_common.assign(specific_returns=specific_pa)
+
+    return strategy_daily_holdings, risk_exposures, perf_attrib
+
+
+def create_perf_attrib_stats(strategy_daily_holdings, risk_exposures,
+                             perf_attrib):
+    """
+    Takes perf attribution data over a period of time and computes annualized
+    alpha, multifactor sharpe, risk exposures.
+    """
+
+    summary = {}
+    specific_returns = perf_attrib['specific_returns']
+    common_returns = perf_attrib['common_returns']
+
+    summary['daily_mf_alpha'] = (
+        specific_returns / strategy_daily_holdings.groupby(['dt']).sum()
+    ).mean()
+
+    summary['annualized_mf_alpha'] =\
+        np.power(summary['daily_mf_alpha'] + 1., 252) - 1.
+
+    summary['mf_sharpe'] = summary['daily_mf_alpha'] * np.sqrt(252) / (
+        specific_returns / strategy_daily_holdings.groupby(['dt']).sum()
+    ).std()
+
+    summary['total_specific_returns'] = specific_returns.sum()
+    summary['total_common_returns'] = common_returns.sum()
+
+    summary['total_returns'] =\
+        summary['total_common_returns'] + summary['total_specific_returns']
+
+    summary = pd.Series(summary)
+    return risk_exposures, summary
+
+
+def show_perf_attrib_stats(perf_attrib_data):
+    """
+    Takes perf attribution data over a period of time, computes stats on it,
+    and displays them using `utils.print_table`.
+    """
+    perf_attrib_stats = create_perf_attrib_stats(perf_attrib_data)
+    for df in perf_attrib_stats:
+        print_table(df)
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -15,7 +15,6 @@
 from __future__ import division
 import pandas as pd
 import numpy as np
-import matplotlib.pyplot as plt
 
 from pyfolio.utils import print_table
 
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -20,24 +20,18 @@ from pyfolio.pos import get_percent_alloc
 from pyfolio.utils import print_table
 
 
-def perf_attrib(factor_loadings, positions, factor_returns, returns):
+def perf_attrib(returns, positions, factor_returns, factor_loadings):
     """
     Does performance attribution given risk info.
 
     Parameters
     ----------
-    factor_loadings : pd.DataFrame
-        Factor loadings for all days in the date range, with date and ticker as
-        index, and factors as columns.
+    returns : pd.Series
+        Returns for each day in the date range.
         - Example:
-                               momentum  reversal
-            dt         ticker
-            2017-01-01 AAPL   -1.592914  0.852830
-                       TLT     0.184864  0.895534
-                       XOM     0.993160  1.149353
-            2017-01-02 AAPL   -0.140009 -0.524952
-                       TLT    -1.066978  0.185435
-                       XOM    -1.798401  0.761549
+            2017-01-01   -0.017098
+            2017-01-02    0.002683
+            2017-01-03   -0.008669
 
     positions: pd.DataFrame
         Daily holdings (in dollars or percentages), indexed by date.
@@ -55,16 +49,21 @@ def perf_attrib(factor_loadings, positions, factor_returns, returns):
             2017-01-01  0.002779 -0.005453
             2017-01-02  0.001096  0.010290
 
-    returns : pd.Series
-        Returns for each day in the date range.
+    factor_loadings : pd.DataFrame
+        Factor loadings for all days in the date range, with date and ticker as
+        index, and factors as columns.
         - Example:
-            2017-01-01   -0.017098
-            2017-01-02    0.002683
-            2017-01-03   -0.008669
+                               momentum  reversal
+            dt         ticker
+            2017-01-01 AAPL   -1.592914  0.852830
+                       TLT     0.184864  0.895534
+                       XOM     0.993160  1.149353
+            2017-01-02 AAPL   -0.140009 -0.524952
+                       TLT    -1.066978  0.185435
+                       XOM    -1.798401  0.761549
 
     Returns
     -------
-
     perf_attribution : pd.DataFrame
         df with factors, common returns, and specific returns as columns,
         and datetimes as index
@@ -91,10 +90,10 @@ def perf_attrib(factor_loadings, positions, factor_returns, returns):
                                'common_returns': common_returns,
                                'specific_returns': specific_returns})
 
-    return pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
+    return  pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
 
 
-def create_perf_attrib_stats(positions, perf_attrib):
+def create_perf_attrib_stats(perf_attrib):
     """
     Takes perf attribution data over a period of time and computes annualized
     multifactor alpha, multifactor sharpe, risk exposures.
@@ -103,15 +102,11 @@ def create_perf_attrib_stats(positions, perf_attrib):
     specific_returns = perf_attrib['specific_returns']
     common_returns = perf_attrib['common_returns']
 
-    daily_multifactor_alpha = (
-        specific_returns / positions.groupby(['dt']).sum()
-    ).mean()
-
     summary['Annual multi-factor alpha'] =\
-        empyrical.annual_return(daily_multifactor_alpha)
+        empyrical.annual_return(specific_returns)
 
-    summary['Multifactor sharpe'] =\
-        empyrical.sharpe_ratio(daily_multifactor_alpha)
+    summary['Multi-factor sharpe'] =\
+        empyrical.sharpe_ratio(specific_returns)
 
     summary['Cumulative specific returns'] =\
         empyrical.cum_returns(specific_returns)
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -14,15 +14,13 @@
 # limitations under the License.
 from __future__ import division
 import pandas as pd
-import numpy as np
 
+import empyrical
+from pyfolio.pos import get_percent_alloc
 from pyfolio.utils import print_table
 
 
-def perf_attrib(factor_loadings,
-                factor_returns,
-                strategy_daily_returns,
-                strategy_daily_holdings):
+def perf_attrib(factor_loadings, positions, factor_returns, returns):
     """
     Does performance attribution given risk info.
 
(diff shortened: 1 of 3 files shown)


vikram-narayan:
@twiecki could you take a look? made a fork for now, may close in favor of a branch PR once I can do that

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+            2017-01-01  0.249087  0.935925        1.185012          1.185012
+            2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
+    """
+
+    risk_exposures = factor_loadings.multiply(strategy_daily_holdings,
+                                              axis='rows')
+    risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
+    perf_attrib_style = risk_exposures_portfolio.multiply(factor_returns)
+
+    common_pa = perf_attrib_style.sum(axis=1)
```
georgh0021: `axis='columns'` for readability?

georgh0021: `common_perf_attrib` and `axis='columns'`? I think we prefer verbosity since it makes for more readable code.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+
+    strategy_daily_returns : pd.DataFrame
+        Returns for each day in the date range.
+        - Example:
+                            AAPL       TLT       XOM
+            2017-01-01 -0.014092 -0.003938 -0.001508
+            2017-01-02  0.011921  0.003039 -0.001546
+
+    strategy_daily_holdings: pd.Series
+        Daily holdings for all days in the date range, indexed by date
```
georgh0021: "Daily holdings in dollars", just to be clear.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+    perf_attrib_style = risk_exposures_portfolio.multiply(factor_returns)
+
+    common_pa = perf_attrib_style.sum(axis=1)
+    total_returns_daily = strategy_daily_returns.sum(axis=1)
+    specific_pa = total_returns_daily - common_pa
+
+    perf_attrib_common = perf_attrib_style.assign(common_returns=common_pa)
+    perf_attrib = perf_attrib_common.assign(specific_returns=specific_pa)
+
+    return strategy_daily_holdings, risk_exposures, perf_attrib
```
georgh0021: Why does this function return `strategy_daily_holdings`? It was passed in as an input.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def create_perf_attrib_stats(strategy_daily_holdings, risk_exposures,
+                             perf_attrib):
```
georgh0021: Why does this function require `risk_exposures` as an input? It just hands it back to you (on line 138). I might be missing some context here.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+                        momentum  reversal  common_returns  specific_returns
+            dt
+            2017-01-01  0.249087  0.935925        1.185012          1.185012
+            2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
+    """
+
+    risk_exposures = factor_loadings.multiply(strategy_daily_holdings,
+                                              axis='rows')
+    risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
+    perf_attrib_style = risk_exposures_portfolio.multiply(factor_returns)
```
georgh0021: `perf_attrib_style` will also do attribution to sectors, right? E.g. as far as pyfolio is concerned, the technology factor is no different from the momentum factor: its just another common factor. So maybe `perf_attrib_style` isn't the best name here?

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+        Returns by factor, with date as index and factors as columns
+        - Example:
+                        momentum  reversal
+            2017-01-01  0.002779 -0.005453
+            2017-01-02  0.001096  0.010290
+
+    strategy_daily_returns : pd.DataFrame
+        Returns for each day in the date range.
+        - Example:
+                            AAPL       TLT       XOM
```
georgh0021: This might not be relevant at all, but I believe TLT and XOM are two of our sector ETFs... @twiecki are we concerned that we mention them in a docstring?

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+    perf_attrib = perf_attrib_common.assign(specific_returns=specific_pa)
+
+    return strategy_daily_holdings, risk_exposures, perf_attrib
+
+
+def create_perf_attrib_stats(strategy_daily_holdings, risk_exposures,
+                             perf_attrib):
+    """
+    Takes perf attribution data over a period of time and computes annualized
+    alpha, multifactor sharpe, risk exposures.
```
georgh0021: Perhaps we should call it the annualized multifactor alpha, to avoid confusion with the alpha that e.g. Zipline currently calculates.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+
+    summary = {}
+    specific_returns = perf_attrib['specific_returns']
+    common_returns = perf_attrib['common_returns']
+
+    summary['daily_mf_alpha'] = (
```
georgh0021: `multifactor` instead of `mf` for verbosity?

pyfolio/perf_attrib.py
twiecki: Technically those can be passed in without problem, but it might be a bit confusing so we can change this to IBM and SBUX or whatever.

pyfolio/perf_attrib.py
twiecki: Good catch. I think holdings need to be in %, too. Otherwise, different capital bases will lead to different results. Elsewhere in the API we expect positions in $ to be passed in, so in code we should call `pyfolio.pos.get_percent_alloc`. If % are passed in, that call should return identity, so it works in both cases.

pyfolio/perf_attrib.py
twiecki: We don't need the daily alpha in the first place, just the annualized one. `annual_multifactor_alpha` is a good name.

twiecki: Actually, there's no reason for the underscores. We can just call it `Annual multi-factor alpha` and `Common returns` etc

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+            2017-01-02  0.001096  0.010290
+
+    strategy_daily_returns : pd.DataFrame
+        Returns for each day in the date range.
+        - Example:
+                            AAPL       TLT       XOM
+            2017-01-01 -0.014092 -0.003938 -0.001508
+            2017-01-02  0.011921  0.003039 -0.001546
+
+    strategy_daily_holdings: pd.Series
```
twiecki: Elsewhere in the API we expect this in wide, so should be wide here too and then do the unstacking in the function.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+
+    summary = {}
+    specific_returns = perf_attrib['specific_returns']
+    common_returns = perf_attrib['common_returns']
+
+    summary['daily_mf_alpha'] = (
+        specific_returns / strategy_daily_holdings.groupby(['dt']).sum()
+    ).mean()
+
+    summary['annualized_mf_alpha'] =\
```
twiecki: instead of calculating manually, use `empyrical.annual_return`.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+        np.power(summary['daily_mf_alpha'] + 1., 252) - 1.
+
+    summary['mf_sharpe'] = summary['daily_mf_alpha'] * np.sqrt(252) / (
```
twiecki: use `empyrical.sharpe_ratio`.

pyfolio/perf_attrib.py
twiecki: Shouldn't these just be the specific returns?

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+        specific_returns / strategy_daily_holdings.groupby(['dt']).sum()
+    ).std()
+
+    summary['total_specific_returns'] = specific_returns.sum()
```
twiecki: rename to `Cumulative specific returns` and call `ep.cum_returns`.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+    summary['total_common_returns'] = common_returns.sum()
```
twiecki: Same.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+
+    summary['total_returns'] =\
```
twiecki: I would just call `cum_returns` on the strategy returns here.

pyfolio/perf_attrib.py
twiecki: Actually, we just need to pass in `returns` which is a `Series` of daily returns of the strategy, not the individual ticker information.

twiecki: Then, you don't need to sum in https://github.com/quantopian/pyfolio/pull/417/files#diff-2b62546bdae188439e7621bb2cc8b2e1R100.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+
+    risk_exposures = factor_loadings.multiply(strategy_daily_holdings,
+                                              axis='rows')
+    risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
+    perf_attrib_style = risk_exposures_portfolio.multiply(factor_returns)
+
+    common_pa = perf_attrib_style.sum(axis=1)
+    total_returns_daily = strategy_daily_returns.sum(axis=1)
+    specific_pa = total_returns_daily - common_pa
```
twiecki: `specific_returns`

pyfolio/perf_attrib.py
twiecki: `common_returns`

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+
+    perf_attrib_common = perf_attrib_style.assign(common_returns=common_pa)
+    perf_attrib = perf_attrib_common.assign(specific_returns=specific_pa)
```
twiecki: Not sure why that's required.

twiecki: I would just do `pd.DataFrame({'returns': returns, 'common_returns': common_returns, 'specific_returns': specific_returns})` and return that.

twiecki:
We also need to think about good tests. I like those where the correct answer is obvious.
Let's create very simple data: 2 days, 2 stocks of holdings, 2 risk-factors. The portfolio returns are `.1, .1`. The risk-exposure should come out to 25% risk-factor 1, 25% risk-factor 2, thus 50% common risk. The common returns should then be `.05, .05` and the specific returns the same.

georgh0021:
@vikram-narayan was just thinking about tests! We could use Rene's example momentum algo... take perhaps the first few days, compute the outputs, and use that as a test?

twiecki:
@georgh0021 Great idea, I think that's a good high-level test. We also want a trivial one in addition to that.

pyfolio/perf_attrib.py
vikram-narayan: renamed things in `summary`,

re specific returns, @twiecki I thought so but one of @georgh0021 's formulas said:
`daily_mf_alpha = (specific_pnl / aums[date_range]).mean()`

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: initially I had wanted `show_perf_attrib_stats` to just take one param, `perf_attrib_data`, so I had `create_perf_attrib_stats` pass the risk exposures to it, but on second look I agree that's unnecessary. I'll remove it here, and add it as a param to `show_perf_attrib_stats`

pyfolio/perf_attrib.py
vikram-narayan: whoops, fixed

pyfolio/perf_attrib.py
vikram-narayan: I assume you mean that along with the other perf attrib stuff? changed it to do that, let me know if that's not what you wanted

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: changed to `perf_attrib_by_factor`

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,148 @@
...
+            2017-01-01  AAPL      71
+                        TLT       93
+                        XOM       10
+            2017-01-02  AAPL      71
+                        TLT       16
+                        XOM       71
+
+    Returns
+    -------
+    exposures : pd.DataFrame
```
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: changed to wide format, called `stack` in `perf_attrib`, and renamed to positions

pyfolio/perf_attrib.py
vikram-narayan: changed to call `get_percent_alloc` inside`perf_attrib`

pyfolio/perf_attrib.py
vikram-narayan: changed this to be a Series called `returns`

pyfolio/perf_attrib.py
vikram-narayan: should `empyrical.annual_return` and `empyrical.sharpe_ratio` get called with `daily_multifactor_alpha` or `specific_returns` as param?

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,134 @@
...
+# limitations under the License.
+from __future__ import division
+import pandas as pd
+
+import empyrical
+from pyfolio.pos import get_percent_alloc
+from pyfolio.utils import print_table
+
+
+def perf_attrib(factor_loadings, positions, factor_returns, returns):
```
twiecki: This should be the order: `returns, positions, factor_returns, factor_loadings`.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,134 @@
...
+            2017-01-03   -0.008669
+
+    Returns
+    -------
+
+    perf_attribution : pd.DataFrame
+        df with factors, common returns, and specific returns as columns,
+        and datetimes as index
+        - Example:
+                        momentum  reversal  common_returns  specific_returns
```
twiecki: Actually, you're right, we also need to return the individual perf-attributed factor-returns from `perf_attrib_by_factor`. You would need to add that below in the `pd.DataFrame` creation.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,134 @@
...
+def create_perf_attrib_stats(positions, perf_attrib):
+    """
+    Takes perf attribution data over a period of time and computes annualized
+    multifactor alpha, multifactor sharpe, risk exposures.
+    """
+    summary = {}
+    specific_returns = perf_attrib['specific_returns']
+    common_returns = perf_attrib['common_returns']
+
+    daily_multifactor_alpha = (
```
twiecki: remove this line and statistic.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,134 @@
...
+        specific_returns / positions.groupby(['dt']).sum()
+    ).mean()
+
+    summary['Annual multi-factor alpha'] =\
+        empyrical.annual_return(daily_multifactor_alpha)
+
+    summary['Multifactor sharpe'] =\
```
twiecki: consistency with `multi-factor`.

twiecki:
See minor fixes, the next important item are tests, as described by me above. Let me know if that case makes sense.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,134 @@
...
+                       TLT     0.184864  0.895534
+                       XOM     0.993160  1.149353
+            2017-01-02 AAPL   -0.140009 -0.524952
+                       TLT    -1.066978  0.185435
+                       XOM    -1.798401  0.761549
+
+    positions: pd.DataFrame
+        Daily holdings (in dollars or percentages), indexed by date.
+        Will be converted to percentages if positions are in dollars.
+        - Example:
```
jzhang18: This example uses daily holdings in dollars.  Since "daily holdings in percentages" is preferred, using an example with daily holdings in percentages may be better to avoid confusion.

vikram-narayan: good point, changed to be in percentages

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: removed this stat

pyfolio/perf_attrib.py
vikram-narayan: fixed

pyfolio/perf_attrib.py
vikram-narayan: concat-ed the returns df and the perf attrib df

vikram-narayan:
@twiecki added a test and made changes based on feedback. would you mind taking another look?

pyfolio/perf_attrib.py
twiecki: @jzhang18 also realized that the portfolio holdings, % or $, require a cash column, otherwise the results will be wrong.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,129 @@
...
+            2017-01-01  0.249087  0.935925        1.185012          1.185012
+            2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
+    """
+    # convert holdings to percentages, and convert positions to long format
+    positions = get_percent_alloc(positions)
+    positions = positions.stack()
+    positions.index = positions.index.set_names(['dt', 'ticker'])
+    risk_exposures = factor_loadings.multiply(positions,
+                                              axis='rows')
+    risk_exposures_portfolio = risk_exposures.groupby(level='dt').sum()
```
twiecki: We should either set the index names of `risk_exposures` ourselves or test that they exist and mark it in the doc string.

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,98 @@
...
+    -------
+    tuple of (returns, factor_returns, positions, factor_loadings)
+    returns : pd.DataFrame
+    factor_returns : pd.DataFrame
+    """
+    start_date = '2017-01-01'
+    periods = 10
+    dts = pd.date_range(start_date, periods=periods)
+    np.random.seed(123)
+    tickers = ['AAPL', 'TLT', 'XOM']
```
twiecki: need `cash`

twiecki:
The importance with cash is the following:
In https://github.com/quantopian/pyfolio/blob/master/pyfolio/pos.py#L22 we divide positions by `position.sum()`. Naively, if you had $100 long and $-100 short you would have 0 in the denominator and get infinity as a result. The odd thing with shorting is that it shows up as cash. Thus, if you were $100 long and $-100 short you would have $100 cash (from the short position) and the denom would be 100 and you'd be %100 long and %100 short. In your test you have a long-only portfolio so the problem does not occur, you can just set the cash column to 0 in that case.

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,98 @@
...
+    return returns, positions, factor_returns, factor_loadings
+
+
+class PerfAttribTestCase(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        (self.returns, self.positions, self.factor_returns,
+         self.factor_loadings) = generate_toy_risk_model_output()
+
+    def test_perf_attrib_simple(self):
```
twiecki: that test looks great!

twiecki:
After adding cash to the doc-string and tests (trivial), as well as finishing the other test, this PR gets my sign-off and is waiting for engineering to review. CC @richafrank

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,98 @@
...
+                                   data=np.random.randn(30, 2))
+
+    return returns, positions, factor_returns, factor_loadings
+
+
+class PerfAttribTestCase(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        (self.returns, self.positions, self.factor_returns,
+         self.factor_loadings) = generate_toy_risk_model_output()
```
yankees714: Is this not used?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def test_perf_attrib_simple(self):
+
+        start_date = '2017-01-01'
+        periods = 2
+        dts = pd.date_range(start_date, periods=periods)
+
+        np.random.seed(123)
```
yankees714: I don't think this is needed?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,98 @@
...
+    Returns
+    -------
+    tuple of (returns, factor_returns, positions, factor_loadings)
+    returns : pd.DataFrame
+    factor_returns : pd.DataFrame
+    """
+    start_date = '2017-01-01'
+    periods = 10
+    dts = pd.date_range(start_date, periods=periods)
+    np.random.seed(123)
```
yankees714: Why is this needed?

vikram-narayan: it makes the random numbers generated by numpy predictable

pyfolio/tests/test_perf_attrib.py
vikram-narayan: yeah, we might use it in the future but it's needed now, I'll remove it

pyfolio/tests/test_perf_attrib.py
vikram-narayan: will remove that. thanks

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,119 @@
...
+                  'risk_factor2': [0.025, 0.025],
+                  'common_returns': [0.05, 0.05],
+                  'specific_returns': [0.05, 0.05],
+                  'total_returns': returns}
+        )
+
+        perf_attrib_output = perf_attrib(returns, positions,
+                                         factor_returns, factor_loadings)
+
+        self.assertTrue(expected_perf_attrib_output.equals(perf_attrib_output))
```
yankees714: Could use `pd.util.testing.assert_frame_equal` here

pyfolio/perf_attrib.py
yankees714: +1 on specifying that cash is required

pyfolio/tests/test_perf_attrib.py
yankees714: In that case I wonder if some sort of encoding scheme might be preferable to random numbers, but I don't think we need to worry about that at this point, since it seems like we're not using this yet?

twiecki: This is common practice for statistical tests. Fine to keep it as is.

pyfolio/tests/test_perf_attrib.py
```diff
@@ -0,0 +1,119 @@
...
+
+        # test long and short positions
+        positions = pd.DataFrame(index=dts,
```
twiecki: I would turn this into a separate test.

twiecki: that test is great!

twiecki:
@richafrank think we can merge this today?

pyfolio/tests/test_perf_attrib.py
vikram-narayan: changed to use `pd.util.testing.assert_frame_equal`

pyfolio/tests/test_perf_attrib.py
vikram-narayan: added `cash` here

pyfolio/perf_attrib.py
vikram-narayan: added cash column and two examples to docstring to make it clearer

twiecki:
@vikram-narayan Merge when ready!

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+    specific_returns = returns - common_returns
+
+    returns_df = pd.DataFrame({'total_returns': returns,
+                               'common_returns': common_returns,
+                               'specific_returns': specific_returns})
+
+    return pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
+
+
+def create_perf_attrib_stats(perf_attrib):
```
yankees714: @vikram-narayan Do we need a test for this?

twiecki: I don't think so, it just calls a few functions that all have unittests in their respective packages.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+    common_returns = perf_attrib['common_returns']
+
+    summary['Annual multi-factor alpha'] =\
+        empyrical.annual_return(specific_returns)
+
+    summary['Multi-factor sharpe'] =\
+        empyrical.sharpe_ratio(specific_returns)
+
+    summary['Cumulative specific returns'] =\
+        empyrical.cum_returns(specific_returns)
```
twiecki: this needs to call `cum_returns_final`

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+    summary['Cumulative common returns'] =\
+        empyrical.cum_returns(common_returns)
```
twiecki: same

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+    summary['Total returns'] =\
+        empyrical.cum_returns(perf_attrib['total_returns'])
+
+    summary = pd.Series(summary)
+    return summary
+
+
+def show_perf_attrib_stats(perf_attrib_data, risk_exposures):
```
twiecki: I think this would be more useful if it would take the same input arguments as `perf_attrib`, call `perf_attrib` and `create_perf_attrib_stats` and display the info. Perhaps alternatively it could work with pre-computed perf_attrib_data.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+    perf_attrib_by_factor = risk_exposures_portfolio.multiply(factor_returns)
+
+    common_returns = perf_attrib_by_factor.sum(axis='columns')
+    specific_returns = returns - common_returns
+
+    returns_df = pd.DataFrame({'total_returns': returns,
+                               'common_returns': common_returns,
+                               'specific_returns': specific_returns})
+
+    return pd.concat([perf_attrib_by_factor, returns_df], axis='columns')
```
twiecki: We should also return `risk_exposures_portfolio` here.

pyfolio/perf_attrib.py
```diff
@@ -0,0 +1,138 @@
...
+            dt
+            2017-01-01  0.249087  0.935925        1.185012          1.185012
+            2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
+    """
+    # convert holdings to percentages, and convert positions to long format
+    positions = get_percent_alloc(positions)
+    # remove cash after normalizing positions
+    del positions['cash']
+    positions = positions.stack()
+    positions.index = positions.index.set_names(['dt', 'ticker'])
```
twiecki: we should also set this for the other passed arguments, like `factor_loadings`.

PR #428, opened 2017-08-24
Fix Fama-French for short time periods

gusgordon:
Use minimum of rolling window and factor returns when constructing DataFrame for FFF. This is significant for algos with less than 8 (not 6!) months of recent data, because the daily factors on French's Dartmouth site are ~2 months out of date (which is where pandas grabs them from): http://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library.html

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -583,7 +583,7 @@ def rolling_fama_french(returns, factor_returns=None,
     factor_returns['const'] = 1
 
     # have NaNs when there is insufficient data to do a regression
-    regression_coeffs = np.empty((rolling_window,
+    regression_coeffs = np.empty((min(rolling_window, len(factor_returns)),
                                   len(factor_returns.columns)))
     regression_coeffs.fill(np.nan)
 


eigenfoo:
@gusgordon we should probably docstring this new behavior. Also, the rolling FF plot says "6-month" in its title, which might no longer be true with this change.

screen shot 2017-08-24 at 9 44 26 am

gusgordon:
It should be the same, this is just so it doesn't error out if you pass it < 6 (or 8, for now) months of returns data.

twiecki:
Hm, I think it's better in that case to not plot anything and just say not enough data for Fama-French. A regression with fewer than 125 data points isn't really stable.

gusgordon:
@twiecki It should be blank with insufficient data as-is. It doesn't do the regression if there is insufficient data, because the list indexing at the top of the for loop will return a blank list if factor returns is smaller than the rolling window. This just makes it so it can create the DataFrame of NaNs with hitting an exception.

twiecki:
Ah, I misread that. Where does it break otherwise?

twiecki:
OK, I see the problem. This is a good fix. A unit-test would be great.

PR #427, opened 2017-08-23
REL pyfolio 0.8.0

eigenfoo:
@twiecki @vikram-narayan I also updated `__init__.py` while I was at it.

Closes #426

PR #434, opened 2017-08-30
ENH check tear sheet inputs

### LINKED ISSUE #211: plot_holding function calcuate the holding with some mistake
the mistake happens in line 345 in the plotting .py file:
```Python
    df_holdings = positions.apply(lambda x: np.sum(x != 0), axis='columns')
```
since ```np.NaN != 0```here
```
 np.sum(x != 0)
```
includes the case when position is NaN, which shoud be ignored when counting holdings. 
also I believe compare a float with zero is not a good practice.

eigenfoo:
Closes #211 

New dependency: `engarde>= 0.4.0` 

I made a new module with a bunch of functions to apply basic sanity checks to `positions`, `returns` and `transactions`. There's a master `check_inputs` function, which will run the appropriate checks on whatever is passed in. At the start of every `create_foo_tear_sheet` function, we just need to call `check_inputs`, and it will raise a special `TearSheetInputError` exception if anything's wrong.

Completely untested and not yet ready for review: I just wanted to kick off a Travis test to see if I broke anything. I'll play around with the example notebooks to see what this looks like when I get the chance later tonight/tomorrow.

*This PR is a bit confusing because it also has my commits from #427. This should stop being a problem once that PR is merged, though.*


twiecki:
Wow, this is incredible. Will review soon!

.bumpversion.cfg
```diff
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = True
-current_version = 0.1
+current_version = 0.8.0
```
twiecki: We don't use this file anymore, can be deleted.

pyfolio/check_inputs.py
```diff
@@ -0,0 +1,127 @@
...
+    - Positions must be floats
+    """
+    try:
+        ed.none_missing(positions)
+    except AssertionError as e:
+        raise TearSheetInputError('''positions has NaNs. engarde raises the
+                                  following: ''' + e)
+
+    if positions.index.tz is None:
+        raise TearSheetInputError('positions index is not timezone-localized')
```
twiecki: would be even better if we say how to fix it.

twiecki:
Seems like we use `engarde` in only a few places, do we really need it?

.bumpversion.cfg
eigenfoo: Will delete on the 0.8.0 PR.

eigenfoo:
I was actually coming around to that conclusion myself: I'll take a look at what `engarde` actually does and see if we can write our own checks.

eigenfoo:
It's been almost a year. Closing this PR for now, maybe we'll revisit it in the future.

twiecki:
Actually I really liked this effort. Nevermind engarde. But we should automatically tz-convert and tell the user.

PR #440, opened 2017-09-05
Perf attribution by time period

vikram-narayan:
this pr adds `perf_attrib_periods` which splits perf attrib calls into different time periods

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -23,7 +23,7 @@ from pyfolio.utils import print_table, set_legend_location, COLORS
 
 
 def perf_attrib(returns, positions, factor_returns, factor_loadings,
-                pos_in_dollars=True):
+                pos_in_dollars=True, periods=None):
     """
     Does performance attribution given risk info.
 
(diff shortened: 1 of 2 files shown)


pyfolio/perf_attrib.py
```diff
@@ -124,6 +124,86 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
+             'time_period3': '2017-01-20'}
+
+    Returns
+    -------
+    perf_attrib_by_period : dict
+        dict keyed by the same keys as `periods`, where the values are the
+        output of `perf_attrib`, i.e., a tuple of
+        (risk_exposures_portfolio, perf_attribution).
+    """
+    if periods is None:
```
twiecki: I wonder if need to add a new function for this. Could we make `perf_attrib` take a `periods` kwarg and call `perf_attrib` recursively?

vikram-narayan: removed `perf_attrib_periods`, and moved that logic into `perf_attrib`

twiecki:
Sorry, I just realized that the result is not changed where we do the slicing. As such, we don't need to push that logic to `perf_attrib` but can just slice the output data as we need it. As such, we should hold off on this PR and wait for the tear-sheet to be pushed where we can add this functionality.

PR #442, opened 2017-09-21
Summary table

vikram-narayan:
add an optional `cost` param to `plot_returns`.  if present, total_returns less cost is plotted.

### DIFF
--- a/pyfolio/bayesian.py
+++ b/pyfolio/bayesian.py
@@ -86,7 +86,7 @@ def model_returns_t_alpha_beta(data, bmark, samples=2000):
                  mu=mu_reg,
                  sd=sigma,
                  observed=y)
-        trace = pm.sample(samples)
+        trace = pm.sample(samples, progressbar=False)
 
     return model, trace
 
@@ -125,7 +125,7 @@ def model_returns_normal(data, samples=500):
             returns.distribution.variance**.5 *
             np.sqrt(252))
 
-        trace = pm.sample(samples)
+        trace = pm.sample(samples, progressbar=False)
     return model, trace
 
 
@@ -167,7 +167,7 @@ def model_returns_t(data, samples=500):
                          returns.distribution.variance**.5 *
                          np.sqrt(252))
 
-        trace = pm.sample(samples)
+        trace = pm.sample(samples, progressbar=False)
     return model, trace
 
 
@@ -252,7 +252,7 @@ def model_best(y1, y2, samples=1000):
                          returns_group2.distribution.variance**.5 *
                          np.sqrt(252))
 
-        trace = pm.sample(samples)
+        trace = pm.sample(samples, progressbar=False)
     return model, trace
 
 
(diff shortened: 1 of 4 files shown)


vikram-narayan:
@jzhang18 mind taking a quick look?

jzhang18:
@vikram-narayan Sure. I will read it asap.

jzhang18:
@vikram-narayan It looks good to me! Thanks!

vikram-narayan:
Gotcha - @gusgordon should I add that to other plots, and the main perf attrib tearsheet? In a different PR @twiecki said
> As such, we don't need to push that logic to perf_attrib but can just slice the output data as we need it. As such, we should hold off on this PR and wait for the tear-sheet to be pushed where we can add this functionality.

gusgordon:
@vikram-narayan Right, so we just run the perf_attrib function as normal, then whatever it outputs we slice on the `live_start_date` to break the data out.

Definitely want it broken out in the summary table (All, In-sample, Out-of-sample), but I'm not sure about the plots. Might be nice to have some sort of marker for the live start date, like a vertical line? I would say whatever you guys think is useful and looks ok.

pyfolio/bayesian.py
```diff
@@ -86,7 +86,7 @@ def model_returns_t_alpha_beta(data, bmark, samples=2000):
                  mu=mu_reg,
                  sd=sigma,
                  observed=y)
-        trace = pm.sample(samples)
+        trace = pm.sample(samples, progressbar=False)
```
twiecki: Seems like that fixed the travis issue. This isn't a proper fix, however. Can you make this a kwarg for the top function that defaults to `True` and only set it to `False` in the unittest?

twiecki: This seemed to have fixed the issue but it's not a proper fix. Can you move this up to be a kwarg of the function, default to `True` and set it to `False` in the unittest?

pyfolio/perf_attrib.py
```diff
@@ -14,10 +14,11 @@
 # limitations under the License.
 from __future__ import division
 
+from collections import OrderedDict
```
yankees714: This should be grouped with the standard lib imports

pyfolio/perf_attrib.py
```diff
@@ -164,46 +169,63 @@ def show_perf_attrib_stats(returns, positions, factor_returns,
         pos_in_dollars=pos_in_dollars,
     )
 
-    perf_attrib_stats = create_perf_attrib_stats(perf_attrib_data)
+    perf_attrib_stats, risk_exposure_stats =\
+        create_perf_attrib_stats(perf_attrib_data, risk_exposures)
```
yankees714: Thoughts on having `perf_attrib()` and `create_perf_attrib_stats()` returning things in the same order for consistency? I.e. changing this to `risk_exposure_stat, perf_attrib_stats = ...`

joshpayne:
would it be good on PRs that involve new (or changed) charts and tables in pyfolio (and alphalens) to include screenshots of them in the PR like we do for UI work?

jbredeche:
Yes it would :)

PR #443, opened 2017-09-26
Support pandas>=0.18.1

yankees714:
For [compatibility with zipline](https://github.com/quantopian/zipline/blob/fbdfaf82410e6e5a9c9e7e4daa738599a6679683/setup.py#L170).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,20 @@
 language: python
 sudo: false
 
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
+python: 
+  - 2.7
+  - 3.5
+  - 3.6
+
+env:
+  - PANDAS_VERSION=0.18.1
+  - PANDAS_VERSION=0.19.2
+  - PANDAS_VERSION=0.20.3
+
+matrix:
+  exclude:
+  - python: 3.6
+    env: PANDAS_VERSION=0.18.1
 
 before_install:
   # We do this conditionally because it saves us some downloading if the
@@ -24,7 +34,7 @@ before_install:
   - cp pyfolio/tests/matplotlibrc .
 
 install:
-  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy flake8 seaborn scikit-learn runipy pytables networkx pandas-datareader matplotlib-tests joblib
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas=$PANDAS_VERSION Cython patsy flake8 seaborn scikit-learn runipy pytables networkx pandas-datareader matplotlib-tests joblib
   - source activate testenv
   - pip install nose_parameterized
   #- pip install --no-deps git+https://github.com/quantopian/zipline
--- a/pyfolio/risk.py
+++ b/pyfolio/risk.py
@@ -12,9 +12,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import numpy as np
-import matplotlib.pyplot as plt
 from collections import OrderedDict
+from functools import partial
+
+import matplotlib.pyplot as plt
+import numpy as np
+
 
 SECTORS = OrderedDict([
     (101, 'Basic Materials'),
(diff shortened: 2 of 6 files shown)


twiecki:
We should probably add this to travis as well.

yankees714:
@twiecki how would I go about doing that?

twiecki:
See e.g. https://github.com/quantopian/empyrical/blob/master/.travis.yml#L4

yankees714:
Ah got it, thanks. Do we want all python/pandas combinations?

twiecki:
Yes.

twiecki:
Need to wait for https://github.com/quantopian/pyfolio/pull/442 to fix the current breakage.

vikram-narayan:
merging #442 once tests pass

twiecki:
@yankees714 Can you rebase from master?

.travis.yml
```diff
@@ -1,10 +1,26 @@
 language: python
 sudo: false
 
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
+matrix:
```
vikram-narayan: is adding time to the travis build something we're worried about?

pyfolio/risk.py
```diff
@@ -432,9 +432,22 @@ def compute_volume_exposures(shares_held, volumes, percentile):
...
     grossed_frac = shares_grossed.divide(volumes)
 
-    longed_threshold = 100*longed_frac.quantile(percentile, axis='columns')
-    shorted_threshold = 100*shorted_frac.quantile(percentile, axis='columns')
-    grossed_threshold = 100*grossed_frac.quantile(percentile, axis='columns')
+    # NOTE: To work around a bug with nan-handling in pandas 0.18,
+    #       drop all-nan rows before the call to `quantile`, and then
+    #       restore them by reindexing the final result. This is fixed
+    #       in pandas 0.19.
+    longed_threshold = 100*longed_frac.dropna(how='all').quantile(
```
vikram-narayan: will this fail flake8 if there aren't spaces around `*`?

yankees714: Apparently not, as I was just leaving this as it was. For some reason thought we were consistently not including the whitespace in this project, but now that I look again it seems like we do include it more often that not. I'll update this and the other occurrences in this file.

.travis.yml
yankees714: Possibly, any thoughts on this @twiecki, @richafrank?

twiecki: I'm not worried about it.

richafrank: Each job in the matrix would still take the same amount of time, so we wouldn't suffer any new timeouts. I definitely think we should test an environment that matches what we test for zipline; the others I'm less concerned about. I think we have 4 workers - What's the wall time for a build now?

Is it possible for us to specify the factor, so travis [expands](https://docs.travis-ci.com/user/customizing-the-build#Build-Matrix) the matrix itself?

yankees714: For the last build on this branch it was 20-30 minutes per job, a bit over an hour total.

yankees714: Also yeah looks like we should be able to have travis do the expansion, I'll take a look at that

richafrank: An hour sounds kinda painful to me, but as consumers, what do you all think?

yankees714:
@vikram-narayan I realized there was an issue with one of the workarounds, so I had to modify it (a391f8a), mind taking another look?

pyfolio/risk.py
```diff
@@ -432,22 +435,26 @@ def compute_volume_exposures(shares_held, volumes, percentile):
...
-    #       drop all-nan rows before the call to `quantile`, and then
-    #       restore them by reindexing the final result. This is fixed
-    #       in pandas 0.19.
-    longed_threshold = 100 * longed_frac.dropna(how='all').quantile(
-        percentile,
+    # NOTE: To work around a bug in `quantile` with nan-handling in
+    #       pandas 0.18, use np.nanpercentile by applying to each row of
+    #       the dataframe. This is fixed in pandas 0.19.
+    #
+    # longed_threshold = 100*longed_frac.quantile(percentile, axis='columns')
```
vikram-narayan: cruft?

vikram-narayan:
actually looks like the travis build failed, is that due to the most recent change?

twiecki:
@vikram-narayan I think conda is still broken.

vikram-narayan:
I think @yankees714 included a workaround for that, it looks like python 3.6 and pandas 0.18.1 are incompatible, from one of the travis builds:
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - pandas 0.18.1* -> numpy 1.10* -> python 2.7* -> openssl 1.0.1*
  - python 3.6*
Use "conda info <package>" to see the dependencies for each package.
```

pyfolio/risk.py
yankees714: I was just leaving these as a reference for when we no longer need to support 0.18, does that sound alright?

vikram-narayan: yup sounds good

yankees714:
Only remaining concern here was build time - it currently stands at 45 minutes which seems tolerable to me. Going to merge, we can cut time later if need be.

twiecki:
Awesome! Should probably release a new version then.

joshpayne:
@twiecki should we align to completion of the first round of perf attribution studies/tearsheets?

twiecki:
@joshpayne I think we're fine with what we have now.

PR #444, opened 2017-09-27
Fixes and enhancements to rolling fama-french regression

twiecki:
`rolling_fama_french` is actually more general, as one can also pass in their own risk-factors. I thus renamed the function (by default it will still load FF). While doing that I also found a couple of other problems that I fixed here.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -184,7 +184,7 @@ def plot_rolling_fama_french(returns,
 
     ax.set_ylabel('Beta')
 
-    rolling_beta = timeseries.rolling_fama_french(
+    rolling_beta = timeseries.rolling_regression(
         returns,
         factor_returns=factor_returns,
         rolling_window=rolling_window)
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -584,22 +584,18 @@ def rolling_correlation(returns, factor_returns=None,
     else:
         factor_returns = factor_returns.copy()
 
-    # have NaNs when there is insufficient data to do a regression
-    regression_coeffs = np.empty((min(rolling_window, len(factor_returns)),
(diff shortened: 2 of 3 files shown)


pyfolio/timeseries.py
```diff
@@ -547,7 +547,7 @@ def rolling_beta(returns, factor_returns,
         return out
 
 
-def rolling_fama_french(returns, factor_returns=None,
+def rolling_correlation(returns, factor_returns=None,
                         rolling_window=APPROX_BDAYS_PER_MONTH * 6):
     """
     Computes rolling Fama-French single factor betas using a multivariate
```
gusgordon: This doc line should be updated too.

gusgordon:
Looks good - I'm not super familiar with how the performance attribution works, but this also computes an alpha. Is there any overlapping functionality between perf_attrib and here that we could combine?

pyfolio/plotting.py
```diff
@@ -184,7 +184,7 @@ def plot_rolling_fama_french(returns,
 
     ax.set_ylabel('Beta')
 
-    rolling_beta = timeseries.rolling_fama_french(
+    rolling_beta = timeseries.rolling_regression(
```
gusgordon: This should also be `rolling_correlation` not `rolling_regression` - missed that.

twiecki: Good catch, thanks.

twiecki: Actually should be `rolling_regression` throughout.

twiecki:
OK, names should match up now.

gusgordon:
@twiecki Ok, good to merge from my point of view, if you think this isn't an issue: https://github.com/quantopian/pyfolio/pull/444#discussion_r141683801

PR #445, opened 2017-09-27
Allow users to provide extra rows in the top table

gusgordon:
This allows users to pass in extra rows (as a `dict` or `OrderedDict`) to display in the perf_stats table. They can pass `header_rows` to `create_full_tear_sheet`, `create_simple_tear_sheet`, `create_returns_tear_sheet`, `show_perf_stats`, and `print_table`. HTML will be generated and injected into the top of the table. Also puts the dates/ranges here, below the rows that the user passes.

Also clean up a few random doc strings.

Examples:
screen shot 2017-09-27 at 1 14 14 pm

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -81,7 +81,7 @@ def create_full_tear_sheet(returns,
                            factor_returns=None,
                            factor_loadings=None,
                            pos_in_dollars=True,
-                           return_fig=False):
+                           header_rows=None):
     """
     Generate a number of tear sheets that are useful
     for analyzing a strategy's performance.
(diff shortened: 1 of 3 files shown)


pyfolio/tears.py
```diff
@@ -81,7 +81,7 @@ def create_full_tear_sheet(returns,
                            factor_returns=None,
                            factor_loadings=None,
                            pos_in_dollars=True,
-                           return_fig=False):
```
twiecki: is that right?

twiecki:
Great feature. Is this off of recent master?

pyfolio/tears.py
gusgordon: I think so, for some reason we were accepting this argument but it wasn't being used anymore. return_fig is only used for the create_tear_sheet functions that produce one figure with multiple plots, like the returns tear sheet (the full tear sheet produces a bunch of figures). I am surprised this got by flake8 in the past.

gusgordon:
Yeah off recent master. I will re-run tests once Travis starts passing on master, I think @vikram-narayan is working on that.

joshpayne:
Cool! I'm curious - what problem is this solving for the pyfolio user?

gusgordon:
@joshpayne we can use this internally to print metadata (code name, revision, etc.) about the algos within the tear sheet. Users can print their name, etc in here if they want to show this tear sheet to people. It also formats the dates at the top using the same function, whereas before they were just strings tacked on.

PR #447, opened 2017-09-28
TST: use conda version 4.3.21 on travis builds

vikram-narayan:
version 4.3.27 was causing tests to fail


twiecki:
What were the errors?

yankees714:
@twiecki See https://travis-ci.org/quantopian/pyfolio/jobs/280493508#L837

twiecki:
It's a conda bug: https://github.com/conda/conda/issues/6030#issuecomment-332791446, I think we should just wait until they resolve it.

yankees714:
Is there any downside to pinning though? This way we'll avoid any regressions/incompatibilities in `latest` like this in the future.

.travis.yml
```diff
@@ -10,25 +10,25 @@ before_install:
...
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
+  # - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
   - cp pyfolio/tests/matplotlibrc .
 
 install:
   - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas Cython patsy flake8 seaborn scikit-learn runipy pytables networkx pandas-datareader matplotlib-tests joblib
```
yankees714: Won't this line install all of these requirements regardless of the constraints? Maybe we should just have a `requirements.txt` that we can pip install, then use that as a constraints file below?

.travis.yml
```diff
@@ -10,25 +10,25 @@ before_install:
...
     else
-      wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+      wget https://repo.continuum.io/miniconda/Miniconda3-4.3.21-Linux-x86_64.sh -O miniconda.sh;
     fi
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
+  # - conda update -q conda
```
yankees714: Cruft?

vikram-narayan: I kept it commented out since I assume we'll want it back whenever conda fixes this issue, but I'll delete it for now

.travis.yml
vikram-narayan: yup that makes sense

twiecki:
@yankees714 We want to stay compatible with recent conda versions, so while this could cause some friction like here I'd rather notice things right away (like the user does) rather than having to remember to update things and then have versions go out-of-date. I'm in favor of closing this as conda is working again.

@richafrank might have thoughts too.

richafrank:
My inclination is to peg the version, so that our merges/shipments are not blocked (and we don't repeat this conversation and temporary workaround) when conda or other vendors have new releases.

FWIW I haven't seen too many requests around newer conda versions not working with zipline, even though we peg the conda version installed on travis/AppVeyor for zipline.

twiecki:
@richafrank Sounds reasonable, I'm fine with that.

.travis.yml
```diff
@@ -20,25 +20,25 @@ before_install:
   # We do this conditionally because it saves us some downloading if the
   # version is the same.
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
-      wget https://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
+      wget https://repo.continuum.io/miniconda/Miniconda2-4.3.21-Linux-x86_64.sh -O miniconda.sh;
```
yankees714: Looks like 4.3.27.1 was released with the fix for this particular issue, should we pin to that instead?

vikram-narayan: just made that change

.travis.yml
```diff
@@ -20,25 +20,25 @@ before_install:
...
   - hash -r
   - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
   - cp pyfolio/tests/matplotlibrc .
 
 install:
-  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas=$PANDAS_VERSION Cython patsy flake8 seaborn scikit-learn runipy pytables networkx pandas-datareader matplotlib-tests joblib
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION pandas=$PANDAS_VERSION pytables networkx matplotlib-tests
```
yankees714: Why don't we pip install pytables, networkx, matplotlib-tests?

.travis.yml
```diff
@@ -20,25 +20,25 @@ before_install:
...
   - source activate testenv
-  - pip install nose_parameterized
+  - pip install -r requirements_test.txt -c travis_constraints.txt
+  - pip install nose_parameterized -c travis_constraints.txt
```
yankees714: Can we just pin this in `requirements_test.txt`?

.travis.yml
```diff
@@ -20,25 +20,25 @@ before_install:
...
   # Useful for debugging any issues with conda
   - conda info -a
   - cp pyfolio/tests/matplotlibrc .
 
 install:
-  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION ipython pyzmq numpy scipy nose matplotlib pandas=$PANDAS_VERSION Cython patsy flake8 seaborn scikit-learn runipy pytables networkx pandas-datareader matplotlib-tests joblib
+  - conda create -q -n testenv --yes python=$TRAVIS_PYTHON_VERSION pandas=$PANDAS_VERSION pytables networkx matplotlib-tests
   - source activate testenv
-  - pip install nose_parameterized
+  - pip install -r requirements_test.txt -c travis_constraints.txt
```
yankees714: Can we combine these into one requirements file?

twiecki:
What's the status of this? I'm still not excited about going more and more out of date with conda but we should probably either merge or close.

PR #448, opened 2017-10-03
MAINT: add warnings for stock/date mismatches

vikram-narayan:
- remove dates for which there are no factor exposures
- remove `positions`  for which there are no factor exposures

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -110,10 +110,12 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
 
         warnings.warn("Could not find factor loadings for the following "
                       "stocks: {}. Ignoring for exposure calculation and "
-                      "performance attribution. Coverage ratio: {}/{}."
+                      "performance attribution. Coverage ratio: {}/{}. "
+                      "Average allocation of missing stocks: {} "
                       .format(list(missing_stocks),
                               num_stocks - len(missing_stocks),
-                              num_stocks))
+                              num_stocks,
+                              positions[missing_stocks].mean()))
 
         positions = positions.drop(missing_stocks, axis='columns')
 
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from __future__ import division
+import warnings
 
 from collections import OrderedDict
 import empyrical as ep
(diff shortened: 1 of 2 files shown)


pyfolio/perf_attrib.py
```diff
@@ -97,6 +98,33 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
             2017-01-01  0.249087  0.935925        1.185012          1.185012
             2017-01-02 -0.003194 -0.400786       -0.403980         -0.403980
     """
+
+    missing_stocks = positions.columns.difference(
+        (factor_loadings.index.levels[1]))
```
yankees714: Extra set of parentheses here I believe (also I generally prefer to put the closing paren on the next line when formatting like this)

vikram-narayan: fixed

pyfolio/perf_attrib.py
```diff
@@ -97,6 +98,37 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
+
+        warnings.warn("Could not find factor loadings for the dates: {}. "
+                      "Truncating date range for performance attribution. "
+                      "Coverage ratio: {}/{}."
+                      .format(list(missing_factor_loadings_index),
+                              len(missing_factor_loadings_index),
+                              len(positions.index)))
+
+        positions = positions.loc[missing_factor_loadings_index.index[0]:
+                                  missing_factor_loadings_index.index[-1]]
```
yankees714: Does this handle the case of values missing in between these two dts?

pyfolio/perf_attrib.py
```diff
@@ -97,6 +98,37 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
     """
+
+    missing_stocks = positions.columns.difference(
+        factor_loadings.index.levels[1]
+    )
+
+    # cash will not be in factor_loadings
+    missing_stocks = missing_stocks.drop('cash')
+
+    if len(missing_stocks) > 0:
```
twiecki: I think it would be good to also compute the average summed allocation (sum holdings over missing stocks and take average over time) of the missing stocks and report that to give a sense of how much will be missing.

pyfolio/perf_attrib.py
```diff
@@ -97,6 +98,37 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
+
+        warnings.warn("Could not find factor loadings for the following "
+                      "stocks: {}. Ignoring for exposure calculation and "
+                      "performance attribution.".format(list(missing_stocks)))
+
+        positions = positions.drop(missing_stocks, axis='columns')
+
+    missing_factor_loadings_index =\
+        positions.index.difference(factor_loadings.index.levels[0])
```
twiecki: Rather than the difference, we want to subset of dts in positions, but not in factor_loadings (i.e. assymetrical).

twiecki:
Also needs a test.

pyfolio/perf_attrib.py
vikram-narayan: yup, difference is symmetrical
```
> pd.Index(factor_loadings.index.get_level_values(1).unique()).difference(positions.columns)
Index([], dtype='object')
> positions.columns.difference(factor_loadings.index.get_level_values(1).unique())
Index([u'TLT', u'cash'], dtype='object')
```

pyfolio/perf_attrib.py
vikram-narayan: added this to the missing stocks warning

pyfolio/perf_attrib.py
vikram-narayan: yup, checked in the test case

twiecki:
This is great! Good to merge from my end.

vikram-narayan:
thanks for the reviews! will squash commits and merge.

PR #453, opened 2017-10-10
TST: perf attrib regression test

vikram-narayan:
test perf attrib with two years of data


twiecki:
LGTM, how did you create these?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -152,3 +152,78 @@ def test_perf_attrib_simple(self):
...
         pd.util.testing.assert_frame_equal(expected_exposures_portfolio,
                                            exposures_portfolio)
+
+    def test_perf_attrib_regression(self):
+
+        positions = pd.read_csv('pyfolio/tests/test_data/positions.csv',
+                                index_col=0, parse_dates=True)
+
+        positions.columns = [int(col) if col != 'cash' else col
+                             for col in positions.columns]
```
yankees714: Why is this needed?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -152,3 +152,78 @@ def test_perf_attrib_simple(self):
...
+
+        pd.util.testing.assert_series_equal(expected_common_returns,
+                                            common_returns,
+                                            check_names=False)
+
+        # since factor loadings are ones, portfolio risk exposures
+        # should be ones
+        pd.util.testing.assert_frame_equal(
+            risk_exposures_portfolio,
+            pd.DataFrame(np.ones(risk_exposures_portfolio.shape),
```
yankees714: There's `np.ones_like` for this use case, if you find that any nicer

pyfolio/tests/test_perf_attrib.py
```diff
@@ -152,3 +152,78 @@ def test_perf_attrib_simple(self):
...
+        pd.util.testing.assert_series_equal(returns,
+                                            common_returns,
+                                            check_names=False)
+
+        self.assertTrue(np.isclose(specific_returns, 0).all())
+
+        # specific and common returns combined should equal total returns
+        pd.util.testing.assert_series_equal(returns,
+                                            combined_returns,
+                                            check_names=False)
```
yankees714: Seems like the two checks above would imply this result?

pyfolio/tests/test_perf_attrib.py
vikram-narayan: the columns of `positions` get read in from this csv as all strings (besides the cash column), but in other dataframes, the columns get read in as ints, so multiplying the dfs together fails. Just standardizing them

pyfolio/tests/test_perf_attrib.py
vikram-narayan: thanks, will use that

pyfolio/tests/test_perf_attrib.py
vikram-narayan: yup they should, the specific returns are all very close to zero, but not quite zero (which was why I used `np.isclose` for that check). I just wanted to add this in as an additional sanity check

PR #455, opened 2017-10-13
BLD: Lower scikit-learn requirement to >=0.16.1

yankees714:
Discussed with @gusgordon, the Quantopian platform is currently using 0.16.1, so we would like to support it if possible.


gusgordon:
Thanks - LGTM if the test passes.

PR #457, opened 2017-10-16
MAINT: update perf attrib summary table

vikram-narayan:
update summary table to be of the format: 
```
Summary statistics:
Annualized Specific Return (Alpha): x%
Annualized Common Return (Beta): y%
Total Annualized Return: A%

Specific Sharpe Ratio: x%/std.dev()
```
annualized cost to be added later on.

cc @joshpayne

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -171,22 +171,24 @@ def create_perf_attrib_stats(perf_attrib, risk_exposures):
     specific_returns = perf_attrib['specific_returns']
     common_returns = perf_attrib['common_returns']
 
-    summary['Annual multi-factor alpha'] =\
+    summary['Annualized Specific Return'] =\
         ep.annual_return(specific_returns)
+    summary['Annualized Common Return'] =\
+        ep.annual_return(common_returns)
+    summary['Total Annualized Return'] =\
+        ep.annual_return(perf_attrib['total_returns'])
 
-    summary['Multi-factor sharpe'] =\
+    summary['Specific Sharpe Ratio'] =\
         ep.sharpe_ratio(specific_returns)
 
-    # empty line between common/specific/total returns
-    summary[' '] = ' '
-    summary['Cumulative specific returns'] =\
+    summary['Cumulative Specific Return'] =\
         ep.cum_returns_final(specific_returns)
-    summary['Cumulative common returns'] =\
+    summary['Cumulative Common Return'] =\
         ep.cum_returns_final(common_returns)
-    summary['Total returns'] =\
-        ep.cum_returns_final(total_returns)
+    summary['Total Returns'] =\
+        ep.cum_returns_final(perf_attrib['total_returns'])
 
-    summary = pd.Series(summary)
+    summary = pd.Series(summary, name='Summary Statistics')
 
     risk_exposure_summary = pd.DataFrame(
         data={'Annualized Return': ep.annual_return(total_returns),
@@ -216,7 +218,11 @@ def show_perf_attrib_stats(returns, positions, factor_returns,
     perf_attrib_stats, risk_exposure_stats =\
         create_perf_attrib_stats(perf_attrib_data, risk_exposures)
 
-    print_table(perf_attrib_stats)
+    print_table(perf_attrib_stats.loc[['Annualized Specific Return',
+                                       'Annualized Common Return',
+                                       'Total Annualized Return']])
+
+    print_table(perf_attrib_stats[['Specific Sharpe Ratio']])
     print_table(risk_exposure_stats)
 
 
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -176,7 +176,7 @@ def create_perf_attrib_stats(perf_attrib, risk_exposures):
     summary['Annualized Common Return'] =\
         ep.annual_return(common_returns)
     summary['Total Annualized Return'] =\
-        ep.annual_return(perf_attrib['total_returns'])
+        ep.annual_return(total_returns)
 
     summary['Specific Sharpe Ratio'] =\
         ep.sharpe_ratio(specific_returns)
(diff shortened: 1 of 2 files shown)


gusgordon:
Thanks. Do we want both cumulative and annual, or are we fine with just annual?

vikram-narayan:
I removed the cumulative stats and replaced them with annual stats since that's what matched Josh's spec, but it might be helpful to have both. @joshpayne do you think we should keep the cumulative stats?

joshpayne:
Personally, I think we should focus the summary on being a summary - so let's not have multiple variations in a summary table. That means making a hard decision on what we want in the summary and if there was one variation that mattered the most, it was annualized after talking it through with Jess.

I'm all for having every variation our heart desires deeper in the tearsheet, as an additional table.

vikram-narayan:
sounds good, in that case I'll keep this as is

gusgordon:
Here is an example of why we would like cumulative returns. We use the summary stats for generating monthly snapshots for the fund. Let's say we are analyzing September, and the portfolio was up 1% in September. We want to be able to see how much of our September returns were idiosyncratic, and how much were attributable to common factors. If we have cumulative returns (which is returns for September, in this case), that's at the top. If we are removing that capability temporarily, that's fine, but it needs to be added back soon because we use it.

twiecki:
I agree with @joshpayne.

@gusgordon that's a good reason, but not sure we need it by default. It could be a separate function you have to call to get it.

gusgordon:
Yeah, or maybe we can just have an additional or optional table at the end that has some extra stats like @joshpayne said?

vikram-narayan:
what about an optional param `include_cumulative` which defaults to `False`, and adds cumulative stats to the summary table if `True`?

pyfolio/perf_attrib.py
```diff
@@ -172,7 +177,9 @@ def show_perf_attrib_stats(returns, positions, factor_returns,
     perf_attrib_stats, risk_exposure_stats =\
         create_perf_attrib_stats(perf_attrib_data, risk_exposures)
 
-    print_table(perf_attrib_stats)
+    print_table(perf_attrib_stats[:perf_attrib_stats.index.get_loc(
+        "Cumulative Specific Return"
```
yankees714: Can we write this using `drop`?
```
perf_attrib_stats.drop(['Cumulative Specific Return', 'Cumulative Common Return', 'Total Returns'])
```
Should be clearer, as well as not dependent on the order of the rows.

vikram-narayan: 👍  fixed

abhijeetkalyan:
@vikram-narayan can we drop the empty row in the middle of the summary table? The UX around it feels strange, since it gives the impression that some data is missing.

vikram-narayan:
@abhijeetkalyan I initially added it since Josh requested that there be a space in the table, and that seemed like the easiest way to do so. @joshpayne are you fine with me removing the empty row?

yankees714:
@vikram-narayan Could we also use multiple calls to `print_table()`? (we could use `.loc[]` to select just the rows we want for each)

vikram-narayan:
@yankees714 good point, changed `show_perf_attrib_stats` to do that instead

yankees714:
Looks good! Should we just post a screenshot of that change?

vikram-narayan:
ah true, will post a screenshot

vikram-narayan:
forgot to post the screenshot, added a name to the risk exposures table: screen shot 2017-10-25 at 4 12 57 pm

joshpayne:
The space was added because the other three lines are logically grouped and the first two add up to the third. So I wanted to keep the other metric separate.

Separately, I find the repeat of "Summary Statistics" in the above screenshot confusing/unnecessary.

abhijeetkalyan:
screen shot 2017-10-26 at 9 56 23 am

This was the summary table I have on the Q research version of perf attribution. @yankees714 @vikram-narayan any idea why the table might be styled differently in both screenshots?

joshpayne:
@abhijeetkalyan pandas version differences is my guess.

yankees714:
https://jupyter-notebook.readthedocs.io/en/stable/changelog.html#table-style (via @ssanderson )

vikram-narayan:
rebased against master, changed the names, and fixed the nan's in the exposure table  (@yankees714 and @abhijeetkalyan )

updated screenshot for `show_perf_attrib_stats`:
screen shot 2017-10-27 at 3 45 17 pm

twiecki:
LGTM!

pyfolio/tests/test_perf_attrib.py
```diff
@@ -257,6 +263,30 @@ def test_perf_attrib_regression(self):
             exposures_summary['Average Risk Factor Exposure']
         )
 
+        cumulative_returns_by_factor = pd.Series(
+            [ep.cum_returns_final(perf_attrib_output[c])
```
yankees714: Is there any way we could write a test for these things using a dataset small enough that we can test against explicit expected values? Otherwise this is kind of just using the same logic in the code and the test, hence the bug

vikram-narayan: Yup definitely agree. I added a simpler test for these in `test_perf_attrib_simple`, where all the returns are specific returns, so these values should all be 0. Let me know if I should add more test coverage here

pyfolio/tests/test_perf_attrib.py
```diff
@@ -155,6 +155,22 @@ def test_perf_attrib_simple(self):
...
+        self.assertEqual(perf_attrib_summary['Annualized Specific Return'],
+                         perf_attrib_summary['Total Annualized Return'])
+
+        self.assertEqual(perf_attrib_summary['Cumulative Specific Return'],
+                         perf_attrib_summary['Total Returns'])
+
+        pd.util.testing.assert_frame_equal(
+            exposures_summary,
+            pd.DataFrame(0.0, index=exposures_summary.index,
+                         columns=exposures_summary.columns)
```
yankees714: Nice, can we just be explicit in specifying `index` and `columns`? Otherwise this will pass for any dataframe of zeros, since we're referencing the columns and index of the df being tested to define the expected

vikram-narayan: explicitly listed index and columns

yankees714:
LGTM!

vikram-narayan:
👍  will squash commits and merge

PR #459, opened 2017-10-19
BUG: exposures should be averaged, not summed

vikram-narayan:


Jstauth:
I think it would be helpful for this table to have a label describing the values. something like "Average Risk Factor Exposure" perhaps

vikram-narayan:
sounds good, added that name to the exposures summary

vikram-narayan:
@twiecki could you take a quick look and let me know if it's ok for me to merge?

vikram-narayan:
👍  will squash commits and merge

PR #463, opened 2017-10-30
MAINT: don't do perf attrib if no factor loadings

vikram-narayan:
If there are no factor loadings for any positions, show a warning and exit

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -105,6 +105,12 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
     # cash will not be in factor_loadings
     num_stocks = len(positions.columns) - 1
     missing_stocks = missing_stocks.drop('cash')
+    num_stocks_covered = num_stocks - len(missing_stocks)
+
+    if num_stocks_covered == 0:
+        raise ValueError("Could not perform performance attribution. "
+                         "No factor loadings were available for this "
+                         "algorithm's positions.")
 
     if len(missing_stocks) > 0:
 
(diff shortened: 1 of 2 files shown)


pyfolio/perf_attrib.py
```diff
@@ -105,6 +105,13 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
     # cash will not be in factor_loadings
     num_stocks = len(positions.columns) - 1
     missing_stocks = missing_stocks.drop('cash')
+    num_stocks_covered = num_stocks - len(missing_stocks)
+
+    if num_stocks_covered == 0:
+        warnings.warn("Could not perform performance attribution. No factor "
+                      "loadings were available for this algorithm's "
+                      "positions.")
+        return None, None
```
yankees714: Do all the downstream consumers of this handle `None` return values?

vikram-narayan: ah true, should I just have consumers return `None` too?

yankees714: I guess it depends on the context. Actually, should this just raise an exception? Doesn't seem like there's anything valuable we can do once we hit this case?

vikram-narayan: I'm okay with that - @abhijeetkalyan had initially wanted a warning instead of raising an exception, you okay with raising an exception here?

abhijeetkalyan: We discussed offline - going with an exception

vikram-narayan: fixed

Issue #462, opened 2017-10-26
Be robust to nonexistant Fama-French data


gusgordon:
The Fama-French data hasn't been updated in a couple months: http://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library.html

If a tearsheet is being run with returns data that is purely outside of the FF coverage range, for example if the algo just has returns in the past two weeks, then the tear sheet will fail with this:

```
/home/jovyan/pyfolio/pyfolio/tears.pyc in create_full_tear_sheet(returns, positions, transactions, market_data, benchmark_rets, slippage, live_start_date, sector_mappings, bayesian, round_trips, estimate_intraday, hide_positions, cone_std, bootstrap, unadjusted_returns, risk, style_factor_panel, sectors, caps, shares_held, volumes, percentile, turnover_denom, set_context, factor_returns, factor_loadings, pos_in_dollars, header_rows)
    198         turnover_denom=turnover_denom,
    199         header_rows=header_rows,
--> 200         set_context=set_context)
    201 
    202     create_interesting_times_tear_sheet(returns,

/home/jovyan/pyfolio/pyfolio/plotting.pyc in call_w_context(*args, **kwargs)
     50         if set_context:
     51             with plotting_context(), axes_style():
---> 52                 return func(*args, **kwargs)
     53         else:
     54             return func(*args, **kwargs)

/home/jovyan/pyfolio/pyfolio/tears.pyc in create_returns_tear_sheet(returns, positions, transactions, live_start_date, cone_std, benchmark_rets, bootstrap, turnover_denom, header_rows, return_fig)
    590 
    591     plotting.plot_rolling_fama_french(
--> 592         returns, ax=ax_rolling_risk)
    593 
    594     # Drawdowns

/home/jovyan/pyfolio/pyfolio/plotting.pyc in plot_rolling_fama_french(returns, factor_returns, rolling_window, legend_loc, ax, **kwargs)
    190         rolling_window=rolling_window)
    191 
--> 192     rolling_beta.plot(alpha=0.7, ax=ax, **kwargs)
    193 
    194     ax.axhline(0.0, color='black')

[...]

PR #464, opened 2017-11-01
Check for NaNs before regressing on Fama-French factors

gusgordon:
Fixes https://github.com/quantopian/pyfolio/issues/462

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -604,7 +604,7 @@ def rolling_regression(returns, factor_returns=None,
             factor_returns_period_dnan = factor_returns_period.dropna()
             reg = linear_model.LinearRegression(fit_intercept=True).fit(
                 factor_returns_period_dnan,
-                returns_period.loc[factor_returns_period_dnan].index)
+                returns_period.loc[factor_returns_period_dnan.index])
             rolling_risk.loc[end, factor_returns.columns] = reg.coef_
             rolling_risk.loc[end, 'alpha'] = reg.intercept_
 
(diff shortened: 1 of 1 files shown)


pyfolio/timeseries.py
```diff
@@ -593,11 +593,12 @@ def rolling_regression(returns, factor_returns=None,
     for beg, end in zip(ret_no_na.index[:-rolling_window],
                         ret_no_na.index[rolling_window:]):
         returns_period = ret_no_na[beg:end]
-        reg = linear_model.LinearRegression(fit_intercept=True).fit(
-            factor_returns.loc[returns_period.index],
-            returns_period)
-        rolling_risk.loc[end, factor_returns.columns] = reg.coef_
-        rolling_risk.loc[end, 'alpha'] = reg.intercept_
+        if ~np.any(factor_returns.loc[returns_period.index].isnull()):
```
twiecki: Can't we just drop individual nans rather then dropping the whole period? Also prefer `not` over `~`.

gusgordon: @twiecki What if there are a large proportion of NaNs? Then we might only do the regression on 2 days of data instead of 120 days. And sure, I'll switch to not.

twiecki: Right, we should add a test that e.g. not more than 10% of values are missing.

pyfolio/timeseries.py
```diff
@@ -593,11 +597,16 @@ def rolling_regression(returns, factor_returns=None,
...
                         ret_no_na.index[rolling_window:]):
         returns_period = ret_no_na[beg:end]
-        reg = linear_model.LinearRegression(fit_intercept=True).fit(
-            factor_returns.loc[returns_period.index],
-            returns_period)
-        rolling_risk.loc[end, factor_returns.columns] = reg.coef_
-        rolling_risk.loc[end, 'alpha'] = reg.intercept_
+        factor_returns_period = factor_returns.loc[returns_period.index]
+
+        if np.all(factor_returns_period.isnull().sum() /
```
twiecki: can just do `..isnull().mean()`, and why do you need `all()`?

Issue #454, opened 2017-10-12
Intermittent test failures in test_tears.py


yankees714:
When the failures occur, these tests always seem to fail:
- `test_create_bayesian_tear_sheet_breakdown`
- `test_create_full_tear_sheet_breakdown`
- `test_create_interesting_times_tear_sheet_breakdown`
- `test_create_returns_tear_sheet_breakdown`
- `test_create_simple_tear_sheet_breakdown`

while these seem to pass:
- `test_create_position_tear_sheet_breakdown`
- `test_create_round_trip_tear_sheet_breakdown`
- `test_create_txn_tear_sheet_breakdown`

cc: @vikram-narayan

PR #467, opened 2017-11-03
Improve warnings before perf attribution

yankees714:
- Adds a new check for algorithms whose turnover is too high
- Updates language for the warning emitted when asses are missing from the risk factor loadings

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -82,7 +82,10 @@ def perf_attrib(returns,
                        XOM    -1.798401  0.761549
 
     transactions : pd.DataFrame, optional
-        Executed trade volumes and fill prices.
+        Executed trade volumes and fill prices. Used to check the turnover of
+        the algorithm. Default is None, in which case the turnover check is
+        skipped.
+
         - One row per trade.
         - Trades on different names that occur at the
           same time will have identical indicies.
@@ -91,8 +94,6 @@ def perf_attrib(returns,
             2004-01-09 12:18:01    483      324.12   'AAPL'
             2004-01-09 12:18:01    122      83.10    'MSFT'
             2004-01-13 14:12:23    -75      340.43   'AAPL'
-        - Used to check the turnover of the algorithm.
-        - Default is None, in which case the turnover check is skipped.
 
     pos_in_dollars : bool
         Flag indicating whether `positions` are in dollars or percentages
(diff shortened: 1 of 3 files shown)


pyfolio/perf_attrib.py
```diff
@@ -114,14 +136,24 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
 
     if len(missing_stocks) > 0:
 
-        warnings.warn("Could not find factor loadings for the following "
```
vikram-narayan: I think you'll need to update the tests as well with the new language?

pyfolio/perf_attrib.py
```diff
@@ -114,14 +136,24 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
+            "accounted for in performance attribution.\n"
+            "\n"
+            "The following assets were missing factor loadings: {}. "
+            "Ignoring for exposure calculation and performance attribution. "
+            "Ratio of assets missing: {}. Average allocation of missing "
+            "assets:\n"
+            "\n"
+            "{}.\n"
+        ).format(
+            list(missing_stocks),
```
vikram-narayan: should we also truncate this list if it's too long?

pyfolio/perf_attrib.py
```diff
@@ -114,14 +136,24 @@ def perf_attrib(returns, positions, factor_returns, factor_loadings,
...
-                      "stocks: {}. Ignoring for exposure calculation and "
-                      "performance attribution. Coverage ratio: {}/{}. "
-                      "Average allocation of missing stocks: {} "
-                      .format(list(missing_stocks),
-                              num_stocks_covered,
-                              num_stocks,
-                              positions[missing_stocks].mean()))
+        missing_stocks_warning_msg = (
+            "Could not determine risk exposures for some of this algorithm's "
+            "positions. PnL from the missing assets will not be properly "
```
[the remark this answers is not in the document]
yankees714: @abhijeetkalyan thoughts on this?

pyfolio/perf_attrib.py
yankees714: Planning to do this in a follow-up PR

twiecki: Same with missing dates, don't want to list all of them but just the period.

abhijeetkalyan: Yup - @twiecki that's another issue that we're tracking

pyfolio/perf_attrib.py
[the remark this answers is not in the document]
abhijeetkalyan: Agreed, returns is clearer

yankees714:
@vikram-narayan Should be good for another look

yankees714:
Failing travis job looked like #454, hopefully the next run will be fine

PR #471, opened 2017-11-07
MAINT: truncate warnings

vikram-narayan:
truncate missing stock/date warnings if there are more than five missing stocks/dates

missing stock warning:
truncated warning missing stocks

missing date warning:
truncated warning missing dates

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -138,24 +138,32 @@ def perf_attrib(returns,
     if len(missing_stocks) > 0:
 
         if len(missing_stocks) > 5:
-            missing_stocks_displayed = list(missing_stocks[:5])
+            missing_stocks_displayed = (
+                " {} assets were missing factor loadings, including: {}"
+            ).format(len(missing_stocks),
+                     list(missing_stocks[:5]))
+            avg_allocation_msg = "selected missing assets"
+
         else:
-            missing_stocks_displayed = list(missing_stocks)
+            missing_stocks_displayed = (
+                "The following assets were missing factor loadings: {}."
+            ).format(list(missing_stocks))
+            avg_allocation_msg = "missing assets"
 
         missing_stocks_warning_msg = (
             "Could not determine risk exposures for some of this algorithm's "
             "positions. Returns from the missing assets will not be properly "
             "accounted for in performance attribution.\n"
             "\n"
-            "The following assets were missing factor loadings: {}. "
+            "{}. "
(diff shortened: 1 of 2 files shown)


pyfolio/perf_attrib.py
```diff
@@ -165,9 +170,17 @@ def perf_attrib(returns,
 
     if len(missing_factor_loadings_index) > 0:
 
+        if len(missing_factor_loadings_index) > 5:
+            missing_dates_displayed = "{}..{}".format(
+                missing_factor_loadings_index[0],
+                missing_factor_loadings_index[-1]
```
yankees714: If we have two missing dates on either end of a large range, this will read as if we're missing the entire range. I'd think we'd want to use this format only if we're actually missing all of those dates. We could also do something like `"Could not find factor loadings for {n} dates (first missing is {YYYY-MM-DD}, last missing is {YYYY-MM-DD})."` cc @abhijeetkalyan

abhijeetkalyan: I haven't looked at the code, but from the screenshot I thought that we'd see the `...` only when the dates are contiguous.  If that's not the case, then I think the message you posted in the comment above is a good one, where we show the first missing and last missing.

pyfolio/perf_attrib.py
```diff
@@ -137,6 +137,11 @@ def perf_attrib(returns,
 
     if len(missing_stocks) > 0:
 
+        if len(missing_stocks) > 5:
+            missing_stocks_displayed = list(missing_stocks[:5])
+        else:
+            missing_stocks_displayed = list(missing_stocks)
```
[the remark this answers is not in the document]
vikram-narayan: fixed this to list the number of missing assets if it's more than five, and changed the average allocation message for that as well

pyfolio/perf_attrib.py
vikram-narayan: fixed to use @yankees714's suggestion, so that we don't have to check whether the missing dates are contiguous or not

vikram-narayan:
@yankees714 @abhijeetkalyan updated the warnings language; they now look like this:

screen shot 2017-11-07 at 6 13 21 pm

screen shot 2017-11-07 at 6 27 05 pm

yankees714:
LGTM

abhijeetkalyan:
Suggested a minor change for the assets case to @vikram-narayan in-person

vikram-narayan:
@abhijeetkalyan new screenshot:
screen shot 2017-11-07 at 7 27 59 pm

PR #472, opened 2017-11-08
MAINT: Clarify daily-ness of daily graphs

abhijeetkalyan:
The chart above these is cumulative returns, but both the factor returns and risk exposures are daily. Makes it clear that these charts are both daily in their titles.


joshpayne:
Love this. Had noticed it before but didn't submit the bug.

PR #475, opened 2017-11-08
MAINT: switch colormap to Paired

vikram-narayan:
tab20 colormap was not available in matplotlib 1.5.1, so switching to Paired, which is: https://matplotlib.org/1.5.1/examples/color/colormaps_reference.html

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -17,6 +17,7 @@ from __future__ import division
 
 import warnings
 
+from itertools import cycle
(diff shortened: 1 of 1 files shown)


vikram-narayan:
cc @abhijeetkalyan

vikram-narayan:
switched to static list from https://sashat.me/2017/01/11/list-of-20-simple-distinct-colors/

vikram-narayan:
ok'd in slack by @yankees714

PR #476, opened 2017-11-08
BUG: get_turnover expects positions in dollars

yankees714:
In perf_attrib() we convert positions to percentages, so be sure to
only use positions in dollars when calculating turnover.

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -203,11 +203,7 @@ def perf_attrib(returns,
         factor_returns = factor_returns.drop(missing_factor_loadings_index,
                                              errors='ignore')
 
-    if pos_in_dollars:
-        # convert holdings to percentages
-        positions = get_percent_alloc(positions)
-
-    if transactions is not None:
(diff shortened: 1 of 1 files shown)


pyfolio/perf_attrib.py
```diff
@@ -203,11 +203,7 @@ def perf_attrib(returns,
         factor_returns = factor_returns.drop(missing_factor_loadings_index,
                                              errors='ignore')
 
-    if pos_in_dollars:
-        # convert holdings to percentages
-        positions = get_percent_alloc(positions)
-
-    if transactions is not None:
+    if transactions is not None and pos_in_dollars:
```
ssanderson: Is it intentional that this disables the warning if `pos_in_dollars` is False?

ssanderson: @abhijeetkalyan @yankees714 ^

yankees714: Yep, this is intentional - currently `get_turnover()` doesn't support positions as percentages, but it probably wouldn't be too hard to add support for that.

PR #479, opened 2017-11-09
BUG: drop alpha when plotting rolling fama french

vikram-narayan:
When alpha is in the call to `rolling_beta.plot`, it doesn't show up, which messes up the ordering of the legend.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -189,6 +189,7 @@ def plot_rolling_fama_french(returns,
         factor_returns=factor_returns,
         rolling_window=rolling_window)
 
+    rolling_beta = rolling_beta[['SMB', 'HML', 'Mom']]
     rolling_beta.plot(alpha=0.7, ax=ax, **kwargs)
 
     ax.axhline(0.0, color='black')
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ install_reqs = [
     'scikit-learn>=0.16.1',
     'seaborn>=0.7.1',
     'pandas-datareader>=0.2',
-    'empyrical>=0.3.2'
+    'empyrical>=0.3.3'
 ]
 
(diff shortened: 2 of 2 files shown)


vikram-narayan:
alpha is all NaNs here. Yup, will do that ^

pyfolio/plotting.py
```diff
@@ -189,6 +189,7 @@ def plot_rolling_fama_french(returns,
         factor_returns=factor_returns,
         rolling_window=rolling_window)
 
+    rolling_beta = rolling_beta[['SMB', 'HML', 'Mom   ']]
```
yankees714: Just to make sure, those spaces are actually in the column name?

vikram-narayan: yup, the `Mom` column has those spaces in it

twiecki: We should probably fix the spaces.

yankees714: 👍

twiecki: So there must be a bad string somewhere.

vikram-narayan: looks like it's in the output of `ep.utils.get_fama_french`, will strip the whitespace from there

richafrank: Should we update it in empyrical then?

twiecki: Yes, this makes it tricky though because there's a version dependency now. Thus, I vote we just drop alpha and side-step the issue (and fix it in empyrical).

vikram-narayan: yup, will fix it in empyrical

twiecki:
Unfortunately this PR leads to an exception if empyrical cached the Fama-French factors: https://github.com/quantopian/empyrical/issues/64.

To make this robust I propose we either drop the alpha here or call `.strip()` here as well.

PR #480, opened 2017-11-10
MAINT: Display return values as percentages in perf attrib tables

yankees714:
Before:
screen shot 2017-11-10 at 12 15 01 pm

After:
screen shot 2017-11-10 at 12 15 26 pm

### DIFF
--- a/pyfolio/utils.py
+++ b/pyfolio/utils.py
@@ -175,7 +175,7 @@ def extract_rets_pos_txn_from_zipline(backtest):
     return returns, positions, transactions
 
 
-def print_table(table, name=None, fmt=None, header_rows=None):
+def print_table(table, name=None, fmt=None, formatters=None, header_rows=None):
     """
     Pretty print a pandas DataFrame.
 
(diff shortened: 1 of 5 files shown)


pyfolio/utils.py
```diff
@@ -192,6 +192,11 @@ def print_table(table, name=None, fmt=None, header_rows=None):
         Formatter to use for displaying table elements.
         E.g. '{0:.2f}%' for displaying 100 as '100.00%'.
         Restores original setting after displaying.
+    formatters : list or dict, optional
```
vikram-narayan: should we combine this and `fmt` above, or rename one of them? it's slightly confusing to have them both

jzhang18:
this looks great!

yankees714:
Update screenshot:
screen shot 2017-11-10 at 3 19 37 pm

pyfolio/utils.py
yankees714: From what I can tell, all the existing call sites that pass `fmt` are passing `'{0:.2f}'` to format float values. `pd.DataFrame.to_html()` accepts a `float_format` arg, so what would you think of  renaming `fmt` to `float_format`, and just forwarding that to `to_html()`? The call sites would then pass `'{0:.2f}.format'`, and the behavior should be the same.

vikram-narayan: 👍

twiecki:
Nice!!

PR #470, opened 2017-11-07
perf attrib legend improvements

vikram-narayan:
- make the end of each line match up with the ordering of the legend
returns_plot

- use `set_color` with colormaps rather than `set_prop_cycle`. Currently using the tab20 colormap: https://matplotlib.org/examples/color/colormaps_reference.html
factor_contribution_plot

PR #483, opened 2017-11-16
Split style sector

vikram-narayan:
- split style and sector factors in main perf attrib tear sheet
- plot cumulative factor returns in `plot_factor_contribution_to_perf`
- screenshot:
separate_style_and_sector_cumulative

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -1531,7 +1531,7 @@ def create_perf_attrib_tear_sheet(returns,
                                        factor_loadings, transactions)
 
     # one section for the returns plot, and for each factor grouping
-    # one section for factor returns, and one for risk exposures 
+    # one section for factor returns, and one for risk exposures
     vertical_sections = 1 + 2 * len(factor_partitions)
     current_section = 0
 
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -44,6 +44,8 @@ except ImportError:
         ImportWarning)
     have_bayesian = False
 
+STYLES = ['momentum', 'size', 'value', 'reversal_short_term', 'volatility']
+
 
 def timer(msg_body, previous_time):
     current_time = time()
@@ -1471,7 +1473,8 @@ def create_perf_attrib_tear_sheet(returns,
                                   factor_loadings,
                                   transactions=None,
                                   pos_in_dollars=True,
-                                  return_fig=False):
+                                  return_fig=False,
+                                  styles=STYLES):
     """
     Generate plots and tables for analyzing a strategy's performance.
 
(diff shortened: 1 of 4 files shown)


pyfolio/tears.py
```diff
@@ -1503,6 +1506,10 @@ def create_perf_attrib_tear_sheet(returns,
 
     return_fig : boolean, optional
         If True, returns the figure that was plotted on.
+
+    styles : list, optional
+        list of styles to show in separate plots. If None, style factors
+        and sectors will be shown in the same plot
```
yankees714: What would you think about making this more generic, and just supporting arbitrary partitions of the factors? Seems like this is less about these being style factors, and more about just splitting up the plots. Maybe something like a dict mapping the factor categories to the factor names?

So for what we're doing here, we could pass:
```
{'sector': ['sector_1', 'sector_2', ... ], 'style': ['style_1', 'style_2', ... ]}
```

jzhang18:
Looks nice! I was wondering if the bottom 4 figures plotted in a subplot with 4 rows and 1 column would be better. The reason is it may need to show more than 10 years data. a subplot with 2 rows and 2 column may be too dense. What do you think?

Not sure if it matters in the plot. There are 11 sector factors and 5 style factors.

I like you put the specific returns in both style factor plot and sector factor plot!

pyfolio/tears.py
vikram-narayan: Yup I like that - I was hesitant to do it with the plots being side by side, but I'll put each plot on its own row as per @jzhang18's feedback, and then pass in a dict instead

vikram-narayan:
@jzhang18 @yankees714 mind taking another look?
updated screenshot:
separate_plots_factor_type

pyfolio/tears.py
```diff
@@ -1513,19 +1530,56 @@ def create_perf_attrib_tear_sheet(returns,
     perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
                                        factor_loadings, transactions)
 
-    vertical_sections = 3
+    vertical_sections = 1 + 2 * len(factor_partitions)
```
yankees714: Mind just adding a comment here that indicates what different sections are? I think it would make it clearer what we want the layout to look like from just reading the code

vikram-narayan: added a clarifying comment

abhijeetkalyan:
🎉

jzhang18:
Looks Good! Just a few suggestions: 1. add xtickers to the first and third figure. 2. let the order of factors in the legend consistent. 3. In this first figure, we could either use "returns" or "cumulative returns" in the legend. Currently, two of them are using "cumulative".

vikram-narayan:
@jzhang18:
1. changed the plots so that each has their own xlabels
2. the legend is meant to line up with the rightmost part of each line in the plot, as per #470
3. I agree that the language should be consistent, going to merge this PR, and then open another one for that stuff

PR #487, opened 2017-11-20
BUG: reversal_short_term -> short_term_reversal

vikram-narayan:


pyfolio/tears.py
```diff
@@ -45,7 +45,7 @@
     have_bayesian = False
 
 FACTOR_PARTITIONS = {
-    'style': ['momentum', 'size', 'value', 'reversal_short_term',
+    'style': ['momentum', 'size', 'value', 'short_term_reversal',
```
ssanderson: It feels a bit strange to me that the default names of these columns are so tightly tied to Quantopian's internal model.  For a non-Quantopian user of the perf attrib tearsheet, this default seems like it's almost certainly unhelpful.  @twiecki @gusgordon what do you think about moving these partition definitions into Q internal code and leaving it up to users to define the logic to split up their perf attribution plots.  This would mean a little bit more work on the quantopian side, but it would mean we're not imposing a very-quantopian-specific default in an open source project, and it would also make it easier for us to change the names of our data columns internally (right now, if we renamed a risk model column internally, I would have to make a change here, which seems strange.)

twiecki: Agreed.

vikram-narayan: Sounds good - I'll move `FACTOR_PARTITIONS` to Q internal code, and change the default to `None` in `create_perf_attrib_tear_sheet`

twiecki: Oh sorry, I misread. I do think we should provide defaults here. They don't necessarily need to be what we use but otherwise I think the use-case is a bit non-obvious. Users could just rename their risk factors to the default naming scheme or change it here.

vikram-narayan: Any thoughts on what the default should be? If Scott's concern is that we don't know what defaults would make sense for non-Quantopian users, then passing in `None` might be easiest, and we could add more detail to the docstring to clarify to users what to pass in.

twiecki: I like the names we have now.

ssanderson: > Oh sorry, I misread. I do think we should provide defaults here. They don't necessarily need to be what we use but otherwise I think the use-case is a bit non-obvious.

I'm not sure how providing quantopian-specific defaults makes the use case more obvious.

As this stands, the most likely outcome of using this function for a non-quantopian user is that they'll end up with empty plots because none of their columns matched a default naming scheme that's tied to our internal risk model implementation.  To fix that problem, the user either has to pass their own partitioning scheme or rename their columns, but renaming is only viable if there's a 1-1 correspondence between their risk model columns and ours, which seems unlikely. So I think the practical outcome for a non-Quantopian user is that they'll have to pass a partitioning scheme no matter what, and by providing our current default we're just making the initial user experience more confusing.

Ultimately I'll defer to your decision on this, but IMO this default amounts to making PyFolio a worse library for non-Quantopian users in exchange for making it more convenient for us to use internally.

twiecki: Thanks @ssanderson. You're right with your assessment but I'm willing to make that trade-off because I predict most users of this feature will be us and users on Quantopian.

twiecki: Maybe we could check if those columns are present and use them if they are, otherwise do nothing.

vikram-narayan: changed this to not plot anything if columns are not present

vikram-narayan: mind taking another look?

ssanderson: @vikram-narayan if we're leaving defaults in this function, then I don't think we should be changing what's currently in source.  Downstream in the Q Research codebase, I think we should be be passing values explicitly here no matter what, and @twiecki's team is likely going to just use the raw files we produce, which are going to have a column name of `reversal_short_term`, which is what was here before.

vikram-narayan: makes sense - should I just close this then?

pyfolio/tears.py
```diff
@@ -1567,10 +1576,13 @@ def create_perf_attrib_tear_sheet(returns,
             current_section += 1
 
         for factor_type, partitions in factor_partitions.iteritems():
+            columns_to_select = portfolio_exposures.columns.intersection(partitions)
+
+            if len(columns_to_select) == 0:
```
ssanderson: This is still going to leave an empty subplot, right?  What's the difference between continuing here and passing an empty dataframe to `plot_risk_exposures`?

ssanderson: Ah, I see, we're removing vertical sections above.

vikram-narayan: I changed the call GridSpec, to match this the number of plots we call, so that there won't be an empty plot:
https://github.com/quantopian/pyfolio/pull/487/files#diff-135fcf845bbceefcd00ccb65ab24afc8R1547

pyfolio/tears.py
```diff
@@ -1536,9 +1536,15 @@ def create_perf_attrib_tear_sheet(returns,
     perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
                                        factor_loadings, transactions)
 
-    # one section for the returns plot, and for each factor grouping
-    # one section for factor returns, and one for risk exposures
-    vertical_sections = 1 + 2 * len(factor_partitions)
+    # only plot partitions for which there are data in perf_attrib_data
+    num_partitions = 0
```
ssanderson: This might be made a little clearer by putting a name on the logic for checking valid partitions, e.g.:
```python
def is_valid_partition(partition_name):
    partition_cols = factor_partitions[partition_name]
    intersection = perf_attrib_data.columns.intersection(partition_cols)
    return bool(len(intersection))

num_partitions = sum(map(is_valid_partition, factor_partitions))
```

pyfolio/tears.py
```diff
@@ -1536,9 +1536,15 @@ def create_perf_attrib_tear_sheet(returns,
...
+    for partitions in factor_partitions.values():
+        if len(perf_attrib_data.columns.intersection(partitions)) > 0:
+            num_partitions += 1
+
+    # one section for the returns plot, and for each factor partition,
+    # one section for factor returns, and one for exposures
+    vertical_sections = 1 + 2 * max(num_partitions, 1)
```
ssanderson: Why is this max(num_partitions, 1)?  If we have no valid partitions, what happens in those two entries?

vikram-narayan: then we just plot all attributed returns on one plot, and all factor exposures on one plot

pyfolio/tears.py
ssanderson: I think the logic for dropping the plots if we don't have data for a partition might still be valuable, though, thinking about it a bit more, I'm not sure who it's benefiting.  For an external user, if their columns match exactly our risk model, then they won't see a difference, and if their columns don't match, then the dropping logic just means that we'll silently ignore some of their data, which will be confusing and hard to debug.

vikram-narayan: yeah, I was thinking about adding a warning each time we drop a plot, but the KeyError they'd hit if we don't have the dropping logic would describe the issue well anyways.

I think I'm going to close this for now, happy to reopen if we decide that logic is helpful in the future

PR #490, opened 2017-11-21
MAINT: add factor_partitions param to full tear sheet

vikram-narayan:


pyfolio/tears.py
```diff
@@ -90,7 +90,8 @@ def create_full_tear_sheet(returns,
                            factor_returns=None,
                            factor_loadings=None,
                            pos_in_dollars=True,
-                           header_rows=None):
+                           header_rows=None,
+                           factor_partitions=None):
```
yankees714: Should this be `factor_partitions=FACTOR_PARTITIONS`?

vikram-narayan: we're currently deciding whether it should be here: https://github.com/quantopian/pyfolio/pull/487

Either way, in the meantime I'm adding this param so that in Q internal code, we can pass `factor_partitions`

vikram-narayan: discussed over slack - going to make this consistent with the call to `create_perf_attrib_tear_sheet`, and then circle back and change it depending on what we decide with #487

PR #494, opened 2017-11-30
BUG: plot cumulative costs in plot_returns

vikram-narayan:
- when plotting cumulative returns, cost should be cumulative as well.
- plot negative cost

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -360,6 +360,7 @@ def plot_returns(perf_attrib_data, cost=None, ax=None):
     returns = perf_attrib_data['total_returns']
     total_returns_label = 'Total returns'
 
+    cumulative_returns_less_costs = returns
     if cost is not None:
         cumulative_returns_less_costs = _cumulative_returns_less_costs(
             returns,
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -360,12 +360,12 @@ def plot_returns(perf_attrib_data, cost=None, ax=None):
     returns = perf_attrib_data['total_returns']
     total_returns_label = 'Total returns'
 
-    cumulative_returns_less_costs = returns
+
+    cumulative_returns_less_costs = _cumulative_returns_less_costs(
+        returns,
+        cost
+    )
     if cost is not None:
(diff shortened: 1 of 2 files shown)


vikram-narayan:
@yankees714 mind taking another look?

pyfolio/perf_attrib.py
```diff
@@ -361,20 +361,25 @@ def plot_returns(perf_attrib_data, cost=None, ax=None):
...
+            returns,
+            cost
+        )
         total_returns_label += ' (adjusted)'
 
     specific_returns = perf_attrib_data['specific_returns']
     common_returns = perf_attrib_data['common_returns']
 
-    ax.plot(ep.cum_returns(returns), color='b', label=total_returns_label)
+    ax.plot(cumulative_returns_less_costs, color='b',
```
yankees714: I think this means we're only plotting returns now if `cost` isn't None?

vikram-narayan: oh yup, fixed that

pyfolio/perf_attrib.py
```diff
@@ -641,3 +646,12 @@ def _stack_positions(positions, pos_in_dollars=True):
...
 
     return positions
+
+
+def _cumulative_returns_less_costs(returns, costs):
+    """
+    Compute cumulative returns, less costs.
+    """
+    if costs is None:
+        costs = 0
```
yankees714: Could just do:
```python
if costs is None:
   return ep.cum_returns(returns)
```

vikram-narayan:
@yankees714 sorry didn't see your request for a test, added that now

pyfolio/tests/test_perf_attrib.py
```diff
@@ -447,3 +451,18 @@ def test_high_turnover_warning(self):
             "This algorithm has relatively high turnover of its positions.",
             str(w[-1].message),
         )
+
+    def test_cumulative_returns_less_costs(self):
+
+        returns = generate_toy_risk_model_output()[0]
```
yankees714: What would you think of hardcoding `returns` to a small array (like we do with `cost`, and then hand deriving the expected values below? Otherwise we're essentially just reimplementing the logic of the function in this test

vikram-narayan: true, hard coded the expected values

Issue #499, opened 2017-12-20
Incorrect Legend for HML in Rolling Fama French Plot


MichaelJMath:
It looks like the HML legend label is incorrect in the `plotting.plot_rolling_fama_french()` function. It currently labels the HML factor as "High Growth". However, this is actually the opposite of what it should be as HML stands for High Book-to-Market minus Low Book-to-Market, in other words, value minus growth. Source: http://mba.tuck.dartmouth.edu/pages/faculty/ken.french/Data_Library/f-f_factors.html

https://github.com/quantopian/pyfolio/blob/3909c49c3875d1be7700f9a7f4c4da60eb957d4c/pyfolio/plotting.py#L196-L199

Maybe the label should simply be "Value (HML)".

twiecki:
Yes, that seems right. Want to do a PR?

PR #501, opened 2017-12-21
Correct plotting.plot_rolling_fama_french() legend

MichaelJMath:
In reference to issue #499, change the legend in the `plotting.plot_rolling_fama_french()` function to reflect HML factor being associated with Value.


twiecki:
This function was removed. Thanks @MichaelJMath!

PR #502, opened 2017-12-21
fix SystemError: <class 'str'> returned a result with an error set

wilsonwang371:
This happens on my particular platform.
python 3.6.3 pandas 0.18.1

SystemError: <class 'str'> returned a result with an error set

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -896,10 +896,10 @@ def get_max_drawdown_underwater(underwater):
 
     valley = np.argmin(underwater)  # end of the period
     # Find first 0
-    peak = underwater[:valley][underwater[:valley] == 0].index[-1]
+    peak = underwater[:valley].loc[underwater[:valley] == 0].index[-1]
     # Find last 0
     try:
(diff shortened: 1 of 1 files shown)


wilsonwang371:
there might cause issue with numpy.ndarray

wilsonwang371:
double checked and it is not possible that the variable could be of type numpy.ndarray. This is only an issue with my pull request on empyrical

pyfolio/timeseries.py
```diff
@@ -896,10 +896,10 @@ def get_max_drawdown_underwater(underwater):
 
     valley = np.argmin(underwater)  # end of the period
     # Find first 0
-    peak = underwater[:valley][underwater[:valley] == 0].index[-1]
+    peak = underwater[:valley].loc[underwater[:valley] == 0].index[-1]
```
eigenfoo: Do you mind elaborating how this fixes the `SystemError` that you got?

wilsonwang371: let me find the root cause and explain it here.

wilsonwang371:
not hitting this error anymore. I guess is it a corner case in one of my previous code. anyway. I am closing this now

Issue #503, opened 2017-12-26
Memory leak?


luca-s:
I believe the `Figures` created in the tear sheet functions should be closed when `return_fig` is `False` otherwise a memory leak is created.

From matplotlib documentation:

The memory required for a Figure is not completely released until the figure is explicitly closed with close(). Deleting all references to the figure, and/or using the window manager to kill the window in which the figure appears on the screen, is not enough, because pyplot maintains internal references until close() is called.

twiecki:
Would they still be displayed in the NB?

luca-s:
If the `Figure` is closed after being shown it will be displayed in the NB. As pyfolio calls `plt.show()` at the end of its tears.* functions it would be possible to close the `Figures` after that statement.

twiecki:
OK, thanks. Want to submit a PR?

luca-s:
Sure

twiecki:
Closed by #504.

PR #504, opened 2018-01-02
BUG: memory leak

luca-s:
Issue #503 

The Figures created in the tear sheet functions should be closed when
'return_fig' is False. The closing of Figures should be done after the
they are shown with plt.show() otherwise they are not displayed.

From matplotlib documentation:

The memory required for a Figure is not completely released until the
figure is explicitly closed with close(). Deleting all references to the
figure, and/or using the window manager to kill the window in which the
figure appears on the screen, is not enough, because pyplot maintains
internal references until close() is called

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -440,6 +440,7 @@ def create_simple_tear_sheet(returns,
         plt.setp(ax.get_xticklabels(), visible=True)
 
     plt.show()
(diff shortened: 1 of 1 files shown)


luca-s:
The test failures are not due to this PR, the HEAD of master branch fails on the same tests

twiecki:
Thanks @luca-s!

PR #514, opened 2018-01-31
MAINT: move perf attrib computations to empyrical

vikram-narayan:
- core performance attribution logic is moved to empyrical (logic to display perf attrib stats is still in pyfolio). [empyrical PR](https://github.com/quantopian/empyrical/pull/76)
- move `get_turnover` to empyrical
- add deprecation warnings to pyfolio perf attrib methods


pyfolio/perf_attrib.py
```diff
@@ -21,13 +21,19 @@
...
 
 PERF_ATTRIB_TURNOVER_THRESHOLD = 0.25
 
+DEPRECATION_WARNING = (
+    "Performance attribution in pyfolio is deprecated. "
+    "Please use empyrical's performance attribution instead."
+)
+
 
+@deprecated(msg=DEPRECATION_WARNING)
```
ssanderson: Instead of keeping two separate copies of this logic, should we call into empyrical here instead?  I don't expect that all of our code that uses this will get migrated immediately, so if we make changes/bugfixes somewhere, it'd be nice to not have to update in two places.

pyfolio/txn.py
```diff
@@ -146,6 +147,8 @@ def adjust_returns_for_slippage(returns, positions, transactions,
     return adjusted_returns
 
 
+@deprecated(msg='pyfolio.get_turnover is deprecated! '
```
ssanderson: Do we need to move turnover? I don't think we're using this downstream anymore.

pyfolio/tears.py
```diff
@@ -1528,16 +1528,16 @@ def create_perf_attrib_tear_sheet(returns,
...
         returns, positions, factor_returns, factor_loadings, transactions,
         pos_in_dollars=pos_in_dollars
     )
 
     display(Markdown("## Performance Relative to Common Risk Factors"))
 
     # aggregate perf attrib stats and show summary table
-    perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
-                                       factor_loadings, transactions)
+    ep.perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
```
ssanderson: I think we only want to move the calculation logic into empyrical.  The plotting logic should still live in pyfolio. cc @richafrank @twiecki does that match your expectation? My general mental model of our project responsibility breakdown is:

- empyrical: Low-level shared risk calculations.
- pyfolio: High-level visualizations of historical portfolio data.
- zipline: Financial simulation, using empyrical definitions for evaluation, outputs feeding into pyfolio.

In particular, I think pyfolio is the only project in this list that should have a matplotlib dependency.

richafrank: Yep, that matches my mental model.

pyfolio/txn.py
vikram-narayan: we use it in perf attrib to determine whether to warn the user, plus I figured it counts as the sort of low level calculation that would make sense in empyrical

ssanderson: Ah, I see, this is a dependency of `perf_attrib`.  I commented in https://github.com/quantopian/empyrical/pull/76#discussion_r165421747 that I'm not sure we want all the validation/warning logic in zipline (where we don't ever expect to have missing or misaligned data, and if we do at least on Quantopian a user won't see the warnings emitted by perf attrib b/c they'll happen at the end of a simulation).  If we just move the performance attribution calculations to empyrical, then I don't think we need to move this, since we're currently only using turnover as a validation step.

ssanderson: The only reason I'm a little hesitant to move this is that the format we use for transactions is currently used only by pyfolio, so right now we could change it and not break much.  Moving this into empyrical means that the format now crosses a project boundary, which makes it harder to change in the future.

pyfolio/tears.py
vikram-narayan: gotcha, I can move this back

vikram-narayan:
updated perf attrib to call empyrical's perf attrib, and kept the warning/alignment logic in pyfolio

pyfolio/perf_attrib.py
```diff
@@ -35,51 +39,11 @@ def perf_attrib(returns,
                 transactions=None,
                 pos_in_dollars=True):
     """
-    Does performance attribution given risk info.
+    Preprocesses inputs and then calls ep.perf_attrib.
 
     Parameters
     ----------
-    returns : pd.Series
```
ssanderson: I would probably still leave these docstrings here.  There's a little bit of a risk that these might get out of date if we make changes in empyrical, but I think it's valuable for a pyfolio user to be able to read these without having to dig through a separate library.  IMO, a user that's using basic pyfolio functionality shouldn't have to know or care about the fact that empyrical exists; it should be mostly an implementation detail of the fact that we need to share code between projects.

pyfolio/perf_attrib.py
```diff
@@ -134,64 +82,17 @@ def perf_attrib(returns,
...
-
-    return (risk_exposures_portfolio,
-            pd.concat([perf_attrib_by_factor, returns_df], axis='columns'))
+    return ep.perf_attrib(returns, positions, factor_returns, factor_loadings)
 
 
 def compute_exposures(positions, factor_loadings, stack_positions=True,
                       pos_in_dollars=True):
     """
-    Compute daily risk factor exposures.
```
ssanderson: I would keep the top-level sentence here and then have the "Normalizes positions" sentence after a newline.  A user reading this docstring shouldn't be expected to know about empyrical or what it does.

pyfolio/perf_attrib.py
```diff
@@ -134,64 +82,17 @@ def perf_attrib(returns,
...
+    Normalizes positions (if necessary) and calls ep.compute_exposures.
 
     Parameters
     ----------
-    positions: pd.DataFrame or pd.Series
```
ssanderson: Same note here about keeping the docstrings.

pyfolio/perf_attrib.py
ssanderson: At some point we might also want to consider something like [`zipline.utils.sharedoc.templated_docstring`](https://github.com/quantopian/zipline/blob/master/zipline/utils/sharedoc.py#L85-L101) for doc chunks that we want to share across libraries.

pyfolio/tests/test_perf_attrib.py
```diff
@@ -75,124 +75,6 @@ def mock_transactions_from_positions(positions):
 
 class PerfAttribTestCase(unittest.TestCase):
 
-    def test_perf_attrib_simple(self):
```
ssanderson: Do we want to keep some portion of this as a regression test so that we don't break this by making an incompatible change in empyrical?  cc @twiecki @richafrank thoughts on this?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -75,124 +75,6 @@ def mock_transactions_from_positions(positions):
...
-                  'risk_factor2': [0.0, 0.0]}
-        )
-
-        pd.util.testing.assert_frame_equal(expected_perf_attrib_output,
-                                           perf_attrib_output)
-
-        pd.util.testing.assert_frame_equal(expected_exposures_portfolio,
-                                           exposures_portfolio)
-
-        perf_attrib_summary, exposures_summary = create_perf_attrib_stats(
```
ssanderson: Is `create_perf_attrib_stats` tested somewhere else?  This seems important to still have coverage for.

vikram-narayan: the [test below this tests `create_perf_attrib_stats`](https://github.com/quantopian/pyfolio/pull/514/files/9f1a616a56ba433978a4281dd51e22ea6cc5af40#diff-e1d9a0a8c573c0639e3b1ffcf47fedb0R116). Should I keep the coverage here?

vikram-narayan:
@ssanderson pushed docstring updates, test failures are due to empyrical version in pyfolio not being updated yet

pyfolio/tests/test_perf_attrib.py
ssanderson: Do you think the test below covers all the functionality we care to test for that function?

pyfolio/tests/test_perf_attrib.py
ssanderson: I think my inclination is to keep these, given that we're still doing a somewhat nontrivial calculation in pyfolio (stacking the positions) before passing positions into empyrical.  Given that we already have these tests here, and that they're not very expensive, I don't see much cost to having them here.

vikram-narayan: added back these tests

richafrank: I buy your point about testing this to ensure what pyfolio is doing to positions before passing them to empyrical works.

PR #517, opened 2018-02-06
Fix #516

### LINKED ISSUE #516: Misleading "Daily Trading Volume" Plot
Right now the DTV plot created by `plot_daily_volume` only plots days where there was indeed trading volume, and connects those plotted days with line making it seem like trading volume linearly changed between days. Leads to a non-sense tear like this:
image

As a fix, I added the below code here: 

```
txn_day_amounts = amounts.groupby(amounts.index).sum()
start = txn_day_amounts.index[0]
end = txn_day_amounts.index[-1]
filled_index = pd.date_range(start, end)
daily_amounts = txn_day_amounts.reindex(filled_index).fillna(0) 
```

here in the file
https://github.com/quantopian/pyfolio/blob/0332bb1600c955a09a581c845fed1a8483b30dc1/pyfolio/txn.py#L105

Could make a PR if that helps.

cyniphile:
Fix #516 plotting issue with daily trading volume tear.

### DIFF
--- a/pyfolio/txn.py
+++ b/pyfolio/txn.py
@@ -108,7 +108,6 @@ def get_txn_vol(transactions):
     end = txn_day_amounts.index[-1]
     filled_index = pd.date_range(start, end)
     daily_amounts = txn_day_amounts.reindex(filled_index).fillna(0) 
-    daily_amounts = amounts.groupby(amounts.index).sum()
     daily_values = values.groupby(values.index).sum()
     daily_amounts.name = "txn_shares"
     daily_values.name = "txn_volume"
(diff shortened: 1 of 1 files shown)


twiecki:
Does that actually work? The sharex sometime causes odd issues. Ideally what we want is to use the same index as on the other plots but I don't we have access to that here. Can you show a plot of the txn plot?

cyniphile:
image

There's the chart. Should probably change, the index of `daily_values` as well for consistency. 

And yes, understand what you are saying. Happy to do it another way if there is one.

twiecki:
I think the correct index is actually important. Imagine someone trying to get the daily average trading volume. As the `date_range` includes weekends this will give the wrong answer. I think cleanest would be to change the API to also require `pos` to be passed in and use the index from there.

twiecki:
Closing due to inactivity.

Issue #509, opened 2018-01-21
[extract_round_trips] ValueError: cannot insert dt, already exists. 


dano2014:
HI. I'm using pyfolio to analyze output from a zipline run. 

import pandas as pd
perf = pd.read_pickle('1594.pickle') # read in perf 
returns, positions, transactions = pf.utils.extract_rets_pos_txn_from_zipline(perf)
rts = pf.round_trips.extract_round_trips(transactions,portfolio_value=positions.sum(axis='columns') / (returns + 1))

The get this error. 

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-43-74bd4b98ea0f> in <module>()
----> 1 rts = pf.round_trips.extract_round_trips(transactions,portfolio_value=positions.sum(axis='columns') / (returns + 1))

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyfolio/round_trips.py in extract_round_trips(transactions, portfolio_value)
    197         rt_returns are the returns in regards to the invested capital
    198         into that partiulcar round-trip.
--> 199     """
    200 
    201     transactions = _groupby_consecutive(transactions)

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pyfolio/round_trips.py in _groupby_consecutive(txn, max_delta)
    122     out = []
    123     for sym, t in txn.groupby('symbol'):
--> 124         t = t.sort_index()
    125         t.index.name = 'dt'
    126         t = t.reset_index()

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pandas/core/frame.py in reset_index(self, level, drop, inplace, col_level, col_fill)
   2959         mapper, index, columns : dict-like or function, optional
   2960             dict-like or functions transformations to apply to
-> 2961             that axis' values. Use either ``mapper`` and ``axis`` to
   2962             specify the axis to target with ``mapper``, or ``index`` and
   2963             ``columns``.

/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/pandas/core/frame.py in insert(self, loc, column, value, allow_duplicates)
   2447             exclude = (exclude,) if exclude is not None else ()
   2448 
-> 2449         selection = tuple(map(frozenset, (include, exclude)))
   2450 
   2451         if not any(selection):

[...]

PR #522, opened 2018-03-01
BUG: #509 remove excessive indexing & adjust portfolio value join & TZ awareness

h55nick:
NOTE: I did not add or adjust tests and have only tested this with my local setup. I do not endorse the merging of this without further testing. Even opening this PR is against my better judgment as a developer.

This fixed: https://github.com/quantopian/pyfolio/issues/509 for my local usage. 

cc/ @twiecki

### DIFF
--- a/pyfolio/round_trips.py
+++ b/pyfolio/round_trips.py
@@ -21,7 +21,7 @@ import pandas as pd
 import numpy as np
 import pytz
 
-from pyfolio.utils import print_table, format_asset
+from utils import print_table, format_asset
 
 PNL_STATS = OrderedDict(
     [('Total profit', lambda x: x.sum()),
--- a/pyfolio/round_trips.py
+++ b/pyfolio/round_trips.py
@@ -19,8 +19,9 @@ from collections import deque, OrderedDict
 
 import pandas as pd
 import numpy as np
+import pytz
 
-from .utils import print_table, format_asset
+from pyfolio.utils import print_table, format_asset
 
 PNL_STATS = OrderedDict(
     [('Total profit', lambda x: x.sum()),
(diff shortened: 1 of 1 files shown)


pyfolio/round_trips.py
```diff
@@ -260,15 +252,21 @@ def extract_round_trips(transactions,
...
-        roundtrips['date'] = roundtrips.close_dt.apply(lambda x:
-                                                       x.replace(hour=0,
-                                                                 minute=0,
-                                                                 second=0))
-
-        tmp = roundtrips.join(pv, on='date', lsuffix='_')
-
-        roundtrips['returns'] = tmp.pnl / tmp.portfolio_value
-        roundtrips = roundtrips.drop('date', axis='columns')
+        roundtrips['date'] = roundtrips.close_dt.apply(
```
twiecki: Does `roundtrips.close_dt.normalize()` work?

h55nick: I'd imagine not, as close_dt would return a Series.
You could `.normalize()` within the lambda? But I still think you'd run into a tz-aware issue as my reading of normalize docs, it does not change the awareness of the TS.

Assuming the problem here is the dep pytz (and generally that this lambda is pretty lacking style in general)

twiecki: OK, there's probably a reason for this pattern here.

twiecki:
What was the specific bug that's fixed here?

twiecki:
Also note that we enforce pep8:
```
pyfolio/round_trips.py:128:80: E501 line too long (85 > 79 characters)
pyfolio/round_trips.py:129:80: E501 line too long (86 > 79 characters)
pyfolio/round_trips.py:132:68: E201 whitespace after '{'
pyfolio/round_trips.py:132:80: E501 line too long (119 > 79 characters)
pyfolio/round_trips.py:266:1: W293 blank line contains whitespace
pyfolio/round_trips.py:268:1: W293 blank line contains whitespace
pyfolio/round_trips.py:411:1: W391 blank line at end of file
```

pyfolio/round_trips.py
h55nick: which pattern exactly?

The only real add as the tz-aware a reordering and then a removal of: `t = t.reset_index()` which is what caused the collision.

twiecki: Just saying, `.normalize()` and `.tz_localize('utc')` should to the same thing, just vectorized.

twiecki:
Closing due to inactivity from submitter.

PR #523, opened 2018-03-27
Fix capacity reporting when creating full tear sheet

traviscross:
When `create_full_tear_sheet` is given market data, it generates the
capacity tear sheet by calling `create_capacity_tear_sheet`.

This was failing.

We were passing to this latter function an incorrectly-named keyword
argument, `daily_vol_limit`.  We probably changed the name of this
argument in the callee during development and didn't update the
caller.

This commit fixes the keyword argument.  Based on the inline constant
used, it appears we meant for this to be
`liquidation_daily_vol_limit`.  This allows capacity reports to be
generated when creating a full tear sheet.

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -240,7 +240,8 @@ def create_full_tear_sheet(returns,
 
             if market_data is not None:
                 create_capacity_tear_sheet(returns, positions, transactions,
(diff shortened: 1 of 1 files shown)


twiecki:
Great, thanks! The travis failure is unrelated to this and it's an obvious bug.

PR #526, opened 2018-04-05
fixed a bug in create_perf_attrib_tear_sheet

jzhang18:
In function `create_perf_attrib_tear_sheet`, it calls `perf_attrib.show_perf_attrib_stats` to show the stats without passing the input `pos_in_dollars`. It would cause problem when positions are in percentage. Fixed it by adding `pos_in_dollars` as input.

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -1553,7 +1553,7 @@ def create_perf_attrib_tear_sheet(returns,
 
     # aggregate perf attrib stats and show summary table
     perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
-                                       factor_loadings, transactions)
+                                       factor_loadings, transactions, pos_in_dollars)
 
     # one section for the returns plot, and for each factor grouping
     # one section for factor returns, and one for risk exposures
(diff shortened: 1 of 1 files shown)


jzhang18:
It shows "All checks have failed", what does it mean...?

pyfolio/tears.py
```diff
@@ -1553,7 +1553,7 @@ def create_perf_attrib_tear_sheet(returns,
 
     # aggregate perf attrib stats and show summary table
     perf_attrib.show_perf_attrib_stats(returns, positions, factor_returns,
-                                       factor_loadings, transactions)
+                                       factor_loadings, transactions, pos_in_dollars)
```
vikram-narayan: you may want to pass this in as `pos_in_dollars=pos_in_dollars`, since `transactions` is also a keyword arg to `show_perf_attrib_stats`

vikram-narayan: looks like flake8 failed as well, so put `pos_in_dollars=pos_in_dollars` on the next line

jzhang18:
@vikram-narayan Just modified the codes based on your comments.

vikram-narayan:
LGTM once travis passes

PR #528, opened 2018-04-13
MAINT: factor_loadings index names should match positions

vikram-narayan:
Make the index names of `factor_loadings` must match `positions` when computing exposures

### DIFF
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -612,6 +612,8 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
+    factor_loadings.index = factor_loadings.index.set_names(['dt', 'ticker'])
+
     return (returns, positions, factor_returns, factor_loadings)
 
 
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -612,6 +612,7 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
+    factor_loadings = factor_loadings.copy()
     factor_loadings.index = factor_loadings.index.set_names(['dt', 'ticker'])
 
     return (returns, positions, factor_returns, factor_loadings)
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -608,8 +608,20 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
-    factor_loadings = factor_loadings.copy(deep=False)
-    factor_loadings.index = factor_loadings.index.set_names(['dt', 'ticker'])
+    # remove any extra data in factor loadings
+    extra_factor_loadings_dates = factor_loadings.index.levels[0].difference(
+        positions.index
+    )
+    if len(extra_factor_loadings_dates):
+        factor_loadings = factor_loadings.drop(extra_factor_loadings_dates)
+
+    extra_factor_loadings_stocks = factor_loadings.index.levels[1].difference(
+        positions.columns
+    )
+    if len(extra_factor_loadings_stocks):
+        factor_loadings = factor_loadings.drop(
+            extra_factor_loadings_stocks, level=1
+        )
 
     return (returns, positions, factor_returns, factor_loadings)
 
--- a/pyfolio/perf_attrib.py
+++ b/pyfolio/perf_attrib.py
@@ -567,10 +567,6 @@ def _align_and_warn(returns,
         factor_loadings.index.get_level_values(0).unique()
     )
 
-    missing_factor_loadings_index = positions.index.difference(
-        factor_loadings.index.get_level_values(0).unique()
-    )
-
     if len(missing_factor_loadings_index) > 0:
 
         if len(missing_factor_loadings_index) > 5:
@@ -612,7 +608,7 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
-    factor_loadings = factor_loadings.copy()
+    factor_loadings = factor_loadings.copy(deep=False)
     factor_loadings.index = factor_loadings.index.set_names(['dt', 'ticker'])
 
     return (returns, positions, factor_returns, factor_loadings)
--- a/pyfolio/tests/test_perf_attrib.py
+++ b/pyfolio/tests/test_perf_attrib.py
@@ -519,8 +519,10 @@ class PerfAttribTestCase(unittest.TestCase):
         ):
             factor_loadings.multiply(positions, axis='rows')
 
+        expected_index = dts.set_names(['dates'])
+
         expected_perf_attrib_output = pd.DataFrame(
-            index=dts,
+            index=expected_index,
             columns=['risk_factor1', 'risk_factor2', 'common_returns',
                      'specific_returns', 'total_returns'],
             data={'risk_factor1': [0.025, 0.025],
@@ -531,7 +533,7 @@ class PerfAttribTestCase(unittest.TestCase):
         )
 
         expected_exposures_portfolio = pd.DataFrame(
-            index=dts,
+            index=expected_index,
             columns=['risk_factor1', 'risk_factor2'],
             data=0.25
         )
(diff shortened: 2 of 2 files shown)


pyfolio/perf_attrib.py
```diff
@@ -612,6 +612,8 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
+    factor_loadings.index = factor_loadings.index.set_names(['dt', 'ticker'])
```
ssanderson: I'm very nervous about mutating the input frame in place like this. Why do we depend specifically on the names of the index levels? If having index level names is absolutely required for this function to work correctly, then IMO we should treat it as an error if they're not what we expect.

vikram-narayan: We want the indexes to align with positions - this lets us pass the output of `get_factor_loadings` to perf attrib directly

ssanderson: I understand that we use the indexes for alignment, but does `perf_attrib` need to depend on the names of the levels of the indexes? As a consumer of this API, I would be surprised (and angry) if the dataframe I passed this function had the level names of its index silently changed.

vikram-narayan: Ah makes sense. Updated this to not change the original input's index names

pyfolio/perf_attrib.py
```diff
@@ -612,6 +612,9 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
+    factor_loadings = factor_loadings.copy()
```
ssanderson: This might be very expensive if `factor_loadings` is large. Where are we actually using the dataframe that we depend explicitly on the names of the index levels? In my experience that's relatively rare (I can't actually think of a case offhand where I've depended on index level names).

vikram-narayan: It's in empyrical, when we compute exposures: `ValueError: cannot join with no level specified and no overlapping names`

ssanderson: Is there a way for us to do that computation without depending on index level name? In general, I don't think we (or at least I) don't think of index level names as being part of the interface of a function. If we require a particular set of names, then we should be documenting that requirement (and I'd also probably argue that we shouldn't be using `ticker`).

ssanderson: Also, it seems like whatever we do for `factor_loadings`, we should be doing for `positions` as well? Is there a reason we're only modifying `factor_loadings` here?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -473,3 +473,82 @@ def test_cumulative_returns_less_costs(self):
...
+
+        tickers = ['stock1', 'stock2']
+        styles = ['risk_factor1', 'risk_factor2']
+
+        returns = pd.Series(data=[0.1, 0.1], index=dts)
+
+        factor_returns = pd.DataFrame(
+            columns=styles,
+            index=dts,
+            data={'risk_factor1': [.1, .1],
```
ssanderson: You can just write `data=0.1` here.

pyfolio/perf_attrib.py
vikram-narayan: Ok, I'll close this PR and open one in empyrical then. For @mmargenot's current alphalens work, should he just change the factor_loadings index name for the time being?

vikram-narayan: We vectorize the positions in pyfolio, and then make this change for positions

ssanderson: w/r/t the performance concern, we might be able to avoid that by passing `deep=False` to the `copy` call here, but I think my question still stands about we're assuming that positions has a particular set of level names but not assuming the same for `factor_loadings`.

vikram-narayan: We set the level names for positions when we vectorize them in `perf_attrib._stack_positions`. Only reason the level names matter is to avoid the `ValueError` that Max saw earlier - but like you said I think we can avoid that by doing the computation without depending on the index level name, which would happen in empyrical

ssanderson: > For @mmargenot's current alphalens work, should he just change the factor_loadings index name for the time being?

I think that depends on what contract we want to provide here. Do we want PyFolio to require users to have to care about things like the level names of their DataFrames (my guess is no).  If not, then I don't think alphalens should have to be changed. Also, if we make that decision, then I think we still want the test in this PR to ensure that we don't care about user-provided level names.

ssanderson: Also, to be clear, "Is there a way for us to do that computation without depending on index level name?" was an actual question.  I'm not sure offhand if there's a straightforward way to do a matrix multiple in pandas on multi-indexed frames with different index level names. So the right fix might very well be for empyrical to be strict and require nicely-formatted input (which in this case would mean requiring that the index level names match), and for PyFolio to do something like what we're doing here (e.g., I think doing a copy with `deep=False` should be fine performance-wise).

vikram-narayan: If I change `compute_exposures` in empyrical to do `groupby(level=0)` rather than `groupby(level='dt').sum()`, perf_attrib will work, but the output of `perf_attrib` will have index names matching the index names of `factor_loadings`.

I think it might be slightly confusing if the outputs of perf_attrib have different index names depending on the input, so I changed copy to use `deep=False` - lmk if you think that's okay.

ssanderson: > I think it might be slightly confusing if the outputs of perf_attrib have different index names depending on the input, so I changed copy to use deep=False - lmk if you think that's okay.

So, if empyrical is doing `groupby(level='dt')`, that means that the interface required by empyrical is that the level 0 must be named 'dt'. That looks like it doesn't match our convention in alphalens, where we appear to use the convention of `'date'`. In another internal Quantopian project, I see similar tables with index level names of `'dates'` and `'sid'`. I also see a variety of different usages just within pyfolio (e.g., I see both `date` and `dt`, as well as both `symbol` and `ticker`).

Given our inconsistent usage across projects, I think my overall preference is for `empyrical` to be agnostic to the names of the indices it receives by using positional level lookups. (i.e., using `groupby(level=0)`), as you suggest.

> I think it might be slightly confusing if the outputs of perf_attrib have different index names depending on the input,

Can you give an example where you think this produces a surprising result? Naively, this doesn't seem like that big of a deal to me, but I'm having trouble imagining a specific example.

vikram-narayan: I meant that if `factor_loadings.index.names` is `['asdf', 'asset']`, the exposures/attributed returns dataframes would have an index name of `asdf`, which might be confusing if you expect the output to be consistently formatted

ssanderson: That doesn't seem that surprising to me. If I hand you a dataframe with `asdf` as the top-level index name, and you do a groupby computation with that dataframe, I wouldn't be surprised to see `asdf` as the name of the resulting index. If anything, I'd be more surprised to get back a dataframe with an index name that's different from anything I passed in.

pyfolio/perf_attrib.py
```diff
@@ -612,6 +608,21 @@ def _align_and_warn(returns,
             )
             warnings.warn(warning_msg)
 
+    # remove any extra data in factor loadings
+    extra_factor_loadings_dates = factor_loadings.index.levels[0].difference(
```
ssanderson: Do we need test coverage for this case now?

pyfolio/tests/test_perf_attrib.py
```diff
@@ -473,3 +473,81 @@ def test_cumulative_returns_less_costs(self):
...
+            index=index,
+            data=0.25
+        )
+        factor_loadings.index =\
+            factor_loadings.index.set_names(['dates', 'asset'])
+
+        # computing factor exposures should fail on its own, but should
+        # pass within perf_attrib because perf_attrib makes index names
+        # match
+        with self.assertRaisesRegexp(
```
ssanderson: I think this will no longer pass with your latest empyrical changes, right?

gusgordon:
Closing this as it's old. Feel free to reopen if this is still relevant.

Issue #534, opened 2018-05-16
MAINT: make pyfolio independent of SPY or FF


eigenfoo:
Currently pyfolio fetches any benchmarks that you do not supply. These happen to be SPY, and the Fama-French factors.

We should relax this: pyfolio should insist that you enter a benchmark, and a set of risk factors. If you do not, it just won't perform any of the appropriate analyses.

https://github.com/quantopian/pyfolio/blob/20060a3045c4176f9a2c417a0cb993b987441885/pyfolio/tears.py#L193-L194

All this means is that lines like these just need to be removed, or raise a `ValueError` to tell the user to decide on a benchmark. Users can use empyrical (e.g. `ep.utils.get_fama_french`) to fetch SPY or FF factors, if they so wish.

This supercedes #495

twiecki:
We shouldn't raise an exception but just skip any analyses that depend on `benchmark` or `fama_french`.

eigenfoo:
Closed by the merged PR.

Issue #532, opened 2018-05-11
Examples don't work


AliAskari:
I have tried to follow up on documents and online tutorials but I cannot seem to make `pyfolio` generate the tear sheats.

Steps to reproduce following [this](https://github.com/quantopian/pyfolio/blob/0ea262f93251f015d70c694d0bd18654bddd7df2/pyfolio/examples/single_stock_example.ipynb) tutorial:

- After trying to fetch the daily returns for any stock, `FB` in the example, 

      stock_rets = pf.utils.get_symbol_rets('FB')

I keep getting the following error from `pandas_datareader`:

    pandas_datareader._utils.RemoteDataError: Unable to read URL: https://finance.google.com/finance/historical?q=FB&startdate=Jan+01%2C+1970&enddate=May+09%2C+2018&output=csv

How can I avoid getting this error or use some locally available data? What should be the format of that data?

twiecki:
Thanks, this is related to https://github.com/quantopian/pyfolio/issues/525.

Issue #530, opened 2018-04-20
Fix travis issues


twiecki:
CC @richafrank

richafrank:
@twiecki Looks like 2 issues: 

1. The tests are erroring in getting data from google finance.
1. test_beta is failing. Looks like differences between successful and failing builds include the versions of numpy and empyrical...

twiecki:
I would suggest to make the code do beta and fama-french analyses optional, only if the right data is passed in. We can then decide if we want to include SPY returns somehow or just rely on the user having access to them if they want to use them.

richafrank:
For the second issue: https://github.com/quantopian/pyfolio/pull/531

twiecki:
Should be fixed by #536.

Issue #525, opened 2018-03-28
yahoo / google problem


aseedb:
I am new to pyfolio. Do I need to install an old version of pandas? 
First I am getting a deprecation error
`ImmediateDeprecationError: 
Yahoo Actions has been immediately deprecated due to large breaks in the API without the
introduction of a stable replacement. Pull Requests to re-enable these data
connectors are welcome.

See https://github.com/pydata/pandas-datareader/issues`

Then while handling this exception, a RemoteDataError is raised. Obviously panda/pyfolio tries to download the data from google finance but that fails.
`RemoteDataError: Unable to read URL: https://finance.google.com/finance/historical?q=FB&startdate=Jan+01%2C+1970&enddate=Mar+27%2C+2018&output=csv`

nemozny:
I am experiencing the same error. You guys are trying to download data from Google and he says

> We're sorry...
> ... but your computer or network may be sending automated queries. To protect our users, we can't process your request right now.

Is there some workaround? How can I switch to a different data source?

Dan733:
Same issue. Currently cannot run a full tearsheet in pyfolio. Does anyone know how to format a custom benchmark to use with `benchmark_rets`?

nemozny:
@Dan733 You have to download your own index and calculate the returns. The format should be the same as in strategy returns.
>date	SPX
>1928-01-04	-0.04

AliAskari:
@nemozny Would you mind explaining more how this should be done?

Do you mean I should also run my strategy on SPX and find the returns of my strat on SPX?

h55nick:
Here is a gist of what I've done to get the working locally (apologizes if it doesn't work as is, copy and pasted from a larger file for brevity)

https://gist.github.com/h55nick/2fa8e9dd71646728565ece4899dd70e1

think the bigger deal here is waiting on master zipline to have the PD upgrade branch merged in so that we can move this to using pd-datareader? The collision between the deps causes some problems.

Hope this helps. Trying to start contributing more!

twiecki:
Thanks @h55nick, want to do a PR?

h55nick:
This PR should do the trick once merged and then bumping the emp dep.

https://github.com/quantopian/empyrical/pull/89

The number of deps in the util.py is slightly worrisome.

twiecki:
I think this is what we should do: https://github.com/quantopian/pyfolio/issues/534

nemozny:
[...]

PR #536, opened 2018-05-18
DEP: Deprecate all data reading functionality via pandas-datareader; ensure independence from SPY and FF

### LINKED ISSUE #499: Incorrect Legend for HML in Rolling Fama French Plot

### LINKED ISSUE #538: MAINT: do not require benchmarks for interesting times tear sheet
#536 makes the interesting times tear sheet require the user to pass a benchmark.

This should not be the case: if a benchmark is passed, plot it, but if not, simply plot cumulative returns for the interesting times. This is because a lot of the "interesting times" don't really have to do with macro-economic/market events.

eigenfoo:
Closes #538 #534 #532 #530 #525 #495 #484 #435 #413 by deprecating all data reading functionality.
Closes #499 by removing rolling Fama-French plots.

Similar to https://github.com/quantopian/empyrical/pull/97

To quote from the README:

As of early 2018, Yahoo Finance has suffered major API breaks with no stable replacement, and the Google Finance API has not been stable since late 2017 [(source)](https://github.com/pydata/pandas-datareader/blob/da18fbd7621d473828d7fa81dfa5e0f9516b6793/README.rst). In recent months it has become a greater and greater strain on the `empyrical` and `pyfolio` development teams to maintain support for fetching data through `pandas-datareader` and other third-party libraries, as these APIs are known to be unstable.

As a result, all `empyrical` (and therefore `pyfolio`, which is a downstream dependency) support for data reading functionality has been deprecated and will be removed in a future version.

Users should beware that the following functions are now deprecated:

- `pyfolio.utils.default_returns_func`
- `pyfolio.utils.get_fama_french`
- `pyfolio.utils.get_returns_cached`
- `pyfolio.utils.get_symbol_returns_from_yahoo`
- `pyfolio.utils.get_treasury_yield`
- `pyfolio.utils.cache_dir`
- `pyfolio.utils.ensure_directory`
- `pyfolio.utils.data_path`
- `pyfolio.utils.load_portfolio_risk_factors`

Users should expect regular failures from the following functions, pending patches to the Yahoo or Google Finance API:

- `pyfolio.utils.default_returns_func`
- `pyfolio.utils.get_symbol_returns_from_yahoo`


eigenfoo:
It seems as if most of pyfolio is already independent of benchmarks. Most functionality in `timeseries.py` and `plotting.py` already check if `benchmark_rets=None`. So this is not as painful a PR as we initially expected, but there's still a lot to do.

eigenfoo:
This PR still isn't ready for merging, btw. Unit tests are done but I will need to do some manual tests to make sure everything is plotting correctly, etc.

README.md
```diff
@@ -80,6 +80,62 @@ If you find a bug, feel free to [open an issue](https://github.com/quantopian/py
...
+   `pandas-datareader`. These include:
+
+   - [Morningstar](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-morningstar)
+   - [Quandl](https://pydata.github.io/pandas-datareader/stable/remote_data.html#remote-data-quandl)
+
+   Please note that you may need to create free accounts with these data
+   providers and receive an API key in order to access data. These API keys
+   should be set as environment variables, or passed as an argument to
+   `pandas-datareader`.
+
```
twiecki: Well written!

pyfolio/tears.py
```diff
@@ -925,12 +913,15 @@ def create_interesting_times_tear_sheet(
     bubble burst, EZB IR, Great Recession (August 2007, March and September
     of 2008, Q1 & Q2 2009), flash crash, April and October 2014.
 
+    benchmark_rets must be passed, as it is meaningless to analyze performance
```
twiecki: Not sure we should make that decision for the user, many events in there are not related to macro market events.

twiecki:
Assigning @richafrank to assign an engineering reviewer.

pyfolio/tears.py
eigenfoo: So we should change the interesting times tear sheet to simply plot cumulative returns during interesting times? That should probably be a separate PR: I'll file an issue for that though.

eigenfoo: Filed under #538

twiecki: My read of the code suggested it was already possible to run it without a benchmark.

eigenfoo: You're right, not much work to do it! Closes #538

eigenfoo: You're right, not too much work to do it.

eigenfoo:
@richafrank is there a status update for this PR?

pyfolio/tears.py
```diff
@@ -86,7 +86,6 @@ def create_full_tear_sheet(returns,
                            volumes=None,
                            percentile=None,
                            turnover_denom='AGB',
-                           set_context=True,
```
ssanderson: Why was this removed?

pyfolio/tears.py
```diff
@@ -497,10 +485,11 @@ def create_returns_tear_sheet(returns, positions=None,
         If True, returns the figure that was plotted on.
     """
 
-    if benchmark_rets is None:
-        benchmark_rets = utils.get_symbol_rets('SPY')
-
-    returns = returns[returns.index > benchmark_rets.index[0]]
+    # If benchmark has been passed, and the strategy's history is longer than
```
ssanderson: It looks like this logic is duplicated in a few places. We might want to consider factoring this out into a helper function named something like `clip_returns_to_benchmark`.

pyfolio/tears.py
```diff
@@ -331,29 +319,34 @@ def create_simple_tear_sheet(returns,
...
-    if benchmark_rets is None:
-        benchmark_rets = utils.get_symbol_rets('SPY')
-
     if (slippage is not None) and (transactions is not None):
         returns = txn.adjust_returns_for_slippage(returns, positions,
                                                   transactions, slippage)
 
     if (positions is not None) and (transactions is not None):
-        vertical_sections = 11
+        vertical_sections = 10
```
ssanderson: Possibly clearer here might be something like:

```python
always_sections = 4
positions_sections = 4 if positions is not None else 0
transactions_sections = 2 if transactions is not None else 0
live_sections = 1 if live_start_date is not None else 0
benchmark_sections = 1 if benchmark_rets is not None else 0

vertical_sections = sum([
    always_sections,
    positions_sections,
    transactions_sections,
    live_sections,
    benchmark_sections,
])
```

pyfolio/tears.py
```diff
@@ -512,16 +501,15 @@ def create_returns_tear_sheet(returns, positions=None,
 
     plotting.show_worst_drawdown_periods(returns)
 
-    # If the strategy's history is longer than the benchmark's, limit strategy
-    if returns.index[0] < benchmark_rets.index[0]:
-        returns = returns[returns.index > benchmark_rets.index[0]]
-
-    vertical_sections = 13
+    vertical_sections = 11
```
ssanderson: Confirming for my own understanding: we removed a section here, right? We used to default to 13. Now we default to 11 and increment by 1 if benchmark_rets was passed. And the removed section corresponds to the deletion of `ax_rolling_risk = plt.subplot(gs[i, :], sharex=ax_rolling_returns)` below?

pyfolio/utils.py
```diff
@@ -402,7 +406,7 @@ def to_series(df):
     return df[df.columns[0]]
 
 
-@deprecated(msg=DEPRECATION_WARNING)
+@deprecated(msg=DATAREADER_DEPRECATION_WARNING)
```
ssanderson: These are already going to raise a warning since they call into deprecated empryical functions, right? Do we want to raise these warnings twice?

pyfolio/utils.py
```diff
@@ -574,6 +577,7 @@ def load_portfolio_risk_factors(filepath_prefix=None, start=None, end=None):
 }
 
 
+@deprecated(msg=DATAREADER_DEPRECATION_WARNING)
```
ssanderson: Do we want to deprecate this functionality? This seems potentially like a nice way to allow a user that has their own reliable source of returns data (including, for example Quantopian), to register their preferred method of returns acquisition without having to explicitly pass returns to every pyfolio function.

I'm not sure how much of a burden that actually is in practice, but we seemed to previously think it was significant enough that we wanted to fetch returns implicitly.

If we decide to leave this, then we probably would want to go back to calling `get_symbol_rets` (maybe renamed to something like `get_default_symbol_rets`) in the places where we used to do so previously, but we'd have the default be to just return None, and we'd have to be prepared to handle a return of None.

If we remove this, we need to make sure that we update anywhere internal on Quantopian where we depend on having registered a function here. In particular, we probably need to make sure we're passing benchmark_rets everywhere that it's desired now.

pyfolio/utils.py
ssanderson: For any of the functions that are just passthroughs to empyrical, I think we can just change our definitions from

```python
def some_func(arg):
    return empyrical.some_func(arg)
```
to just

```python
some_func = empyrical.some_func
```

pyfolio/utils.py
```diff
@@ -49,10 +49,14 @@
...
-                       "use e.g. empyrical.utils.get_symbol_rets() instead "
-                       "of pyfolio.utils.get_symbol_rets()")
+DATAREADER_DEPRECATION_WARNING = \
+        ("Yahoo and Google Finance have suffered large API breaks with no "
+         "stable replacement. As a result, any data reading functionality "
+         "in empyrical has been deprecated and will be removed in a future "
+         "version."
+         ""
+         "Please use empyrical in the Quantopian Research environment, or "
+         "supply your own data. See README.md for more details.")
```
ssanderson: Should we just import this warning message from empyrical?

pyfolio/timeseries.py
```diff
@@ -547,24 +547,21 @@ def rolling_beta(returns, factor_returns,
...
+def rolling_regression(returns, factor_returns,
                        rolling_window=APPROX_BDAYS_PER_MONTH * 6,
                        nan_threshold=0.1):
     """
-    Computes rolling Fama-French single factor betas using a multivariate
-    linear regression (separate linear regressions is problematic because
-    the Fama-French factors are confounded).
-
-    Specifically, returns rolling betas to SMB, HML, and UMD.
+    Computes rolling single factor betas using a multivariate linear regression
```
ssanderson: I'm a bit confused by the expression "single factor betas using a multivariate linear regression". @twiecki is that accurate terminology? Naively I would have expected that "single factor beta" implies a univariate regression.

pyfolio/timeseries.py
```diff
@@ -547,24 +547,21 @@ def rolling_beta(returns, factor_returns,
         return out
 
 
-def rolling_regression(returns, factor_returns=None,
+def rolling_regression(returns, factor_returns,
```
ssanderson: Are there still call-sites of this somewhere in PyFolio? I didn't see any after a quick scan.

pyfolio/utils.py
eigenfoo: Perhaps we should wait until https://github.com/quantopian/empyrical/pull/97 is at least merged before we do?

pyfolio/timeseries.py
eigenfoo: No, but it's possible that users (e.g. Quantopian's research team) depend on this function. In my mind, it is difficult to justify moving this function into, say, `empyrical`, as it is not quite a financial metric, and this sort of calculation usually only arises when analyzing portfolio performance.

pyfolio/plotting.py
```diff
@@ -619,7 +552,7 @@ def show_perf_stats(returns, factor_returns, positions=None,
     returns : pd.Series
         Daily returns of the strategy, noncumulative.
          - See full explanation in tears.create_full_tear_sheet.
-    factor_returns : pd.Series
+    factor_returns : pd.Series, optional
         Daily noncumulative returns of the benchmark.
```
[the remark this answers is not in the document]
eigenfoo: In my mind, there isn't much ambiguity in this docstring: it's the daily returns of some benchmark factor (possibly risk factor). E.g. it could the returns of SPY, or the returns of any of the Fama French factors, etc.

Perhaps "benchmark factor" would be more explicit? I've changed the docstring to reflect that.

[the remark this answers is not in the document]
eigenfoo: With some searching, I'm realizing that the same parameter has very spotty docstrings in different functions. I'm changing all functions to be this same docstring:

```
    factor_returns : pd.Series
         Daily noncumulative returns of the benchmark factor to which betas are
         computed. Usually a benchmark such as market returns.
         - This is in the same style as returns.
```

pyfolio/plotting.py
```diff
@@ -1005,7 +938,7 @@ def plot_rolling_volatility(returns, factor_returns=None,
 
     ax.set_ylabel('Volatility')
     ax.set_xlabel('')
-    if factor_returns.empty:
```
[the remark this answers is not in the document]
eigenfoo: @ssanderson I'm not sure I see how this function would crash if `factor_returns` was `None`. Could you point out a line number?

pyfolio/tears.py
eigenfoo: It was a useless parameter: as far as I could tell, it was passed around to all `create_x_tear_sheet` functions, which did nothing with it. Perhaps `seaborn` makes use of it, but I don't see how it would... @twiecki can you confirm that it's safe to remove this?

pyfolio/utils.py
eigenfoo: Good point; refactored accordingly.

pyfolio/utils.py
eigenfoo: @ssanderson Thinking about it, I agree that this functionality might still be useful, and we shouldn't deprecate it. However, I'm not sure what you mean by "go back to calling `get_symbol_rets`". We currently do not call this function at all in `pyfolio`.

pyfolio/tears.py
eigenfoo: Added `utils.clip_returns_to_benchmark`. This will require a unit test.

pyfolio/tears.py
eigenfoo: That's correct!

pyfolio/utils.py
eigenfoo: Turns out that after dealing with the rest of the feedback, nothing in pyfolio ends up being deprecated at all. So this is kind of a refactor PR?

eigenfoo:
@twiecki finished with engineering feedback; ready for another pass.

twiecki:
Thanks @eigenfoo. @ssanderson can you check if you feedback was appropriately incorporated?

pyfolio/plotting.py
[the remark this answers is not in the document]
ssanderson: On this line, we're accessing `factor_returns.empty`, and `NoneType` doesn't have an `empty` attribute.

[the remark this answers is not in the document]
eigenfoo: Yes, which is why I changed it to `if factor_returns is None`; there's no more reference to the `empty` attribute` anywhere in this function. Am I missing something?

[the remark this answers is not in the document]
ssanderson: @eigenfoo my point here was that, since `factor_returns` was previously required for this function to work, we might want to just consider making it actually a required argument and not call this function at all if we don't have a benchmark. Whether or not that makes sense depends on how useful we think this plot is without access to the benchmark.

pyfolio/tears.py
ssanderson: @eigenfoo it looks to me like this parameter is consumed by the `@customize` decorator in `pyfolio.plotting`:

```python
def customize(func):
    """
    Decorator to set plotting context and axes style during function call.
    """
    @wraps(func)
    def call_w_context(*args, **kwargs):
        set_context = kwargs.pop('set_context', True)
        if set_context:
            with plotting_context(), axes_style():
                return func(*args, **kwargs)
        else:
            return func(*args, **kwargs)
    return call_w_context
```

That decorator is applied to most of the `create_*_tear_sheet` methods.

In general, this strikes me as a scary pattern, because it would be very easy for us to accidentally forget to pass `set_context` somewhere, and we don't have any test coverage enforcing that we handle that parameter correctly. I think trying to fix that problem is probably outside the scope of this PR though. At the very least, I'd revert the removal of that parameter for now.

ssanderson: (In point of fact, it looks like we're failing to pass set_context to several of the tear sheet calls here...)

pyfolio/plotting.py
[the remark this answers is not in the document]
ssanderson: That docstring is definitely an improvement. 

My general objection to this parameter name is that `factor_returns` has ambiguous plurality. Is it expected to be the returns of a single factor (in which case, what factor is it?), or is it a dataframe of factor returns ala the Q risk model? Your updated docstring answers this question for me, but ideally we would have a name here that was more self-describing to begin with. (Changing the name shouldn't be in the scope of this PR though).

pyfolio/utils.py
ssanderson: @eigenfoo on pyfolio master I see at least 5 call-sites of this in the library, plus several in the example notebooks:

```
git --no-pager grep -n -e get_symbol_rets -- 
Binary file pyfolio/examples/bayesian.ipynb matches
Binary file pyfolio/examples/fama_french_benchmark.ipynb matches
Binary file pyfolio/examples/pyfolio_talk_slides.ipynb matches
Binary file pyfolio/examples/single_stock_example.ipynb matches
pyfolio/tears.py:193:        benchmark_rets = utils.get_symbol_rets('SPY')
pyfolio/tears.py:342:        benchmark_rets = utils.get_symbol_rets('SPY')
pyfolio/tears.py:501:        benchmark_rets = utils.get_symbol_rets('SPY')
pyfolio/tears.py:956:        benchmark_rets = utils.get_symbol_rets('SPY')
pyfolio/tears.py:1149:            utils.get_symbol_rets('SPY',
```

pyfolio/tests/test_tears.py
```diff
@@ -116,7 +118,9 @@ def test_create_round_trip_tear_sheet_breakdown(self, kwargs):
     @cleanup
     def test_create_interesting_times_tear_sheet_breakdown(self,
                                                            kwargs):
+        # FIXME needs benchmark_rets
```
ssanderson: Does this still need to be fixed?

pyfolio/utils.py
```diff
@@ -381,190 +374,69 @@ def estimate_intraday(returns, positions, transactions, EOD_hour=23):
...
+default_returns_func = empyrical.utils.default_returns_func
+get_fama_french = empyrical.utils.get_fama_french
+get_returns_cached = empyrical.utils.get_returns_cached
+get_symbol_returns_from_yahoo = empyrical.utils.get_symbol_returns_from_yahoo
+get_treasury_yield = empyrical.utils.get_treasury_yield
+get_utc_timestamp = empyrical.utils.get_utc_timestamp
+cache_dir = empyrical.utils.cache_dir
+ensure_directory = empyrical.utils.ensure_directory
+data_path = empyrical.utils.data_path
+_1_bday_ago = empyrical.utils._1_bday_ago
```
ssanderson: This looks unused in pyfolio. Do we think external consumers were depending on this?

pyfolio/tears.py
eigenfoo: Reverted. We'll definitely need to revisit this in a later PR.

ssanderson: Can you create an issue to track the fact that we're passing this inconsistently?

pyfolio/utils.py
eigenfoo: I couldn't say; I included it just in case though.

pyfolio/tests/test_tears.py
eigenfoo: Nope!

pyfolio/utils.py
eigenfoo: @ssanderson all 5 of those calls have been removed in this PR.

The example notebooks is not something I've thought of, though. It's out of scope for this PR, but we'll need to revisit it later.

pyfolio/utils.py
```diff
@@ -381,190 +374,69 @@ def estimate_intraday(returns, positions, transactions, EOD_hour=23):
     return corrected_positions
 
 
-def to_utc(df):
-    """
-    For use in tests; applied UTC timestamp to DataFrame.
+def clip_returns_to_benchmark(rets, benchmark_rets):
     """
+    If the returns is longer than the benchmark's, limit strategy returns.
```
ssanderson: This description is a bit ambiguous. I would naively interpret "longer than" to mean "has more observations than", but it looks to me like the actual behavior here is that we remove any values from `rets` that are older than the oldest observation in `benchmark_rets`. I also wouldn't know what "limit" means without more context.

I would describe the currently-implemented behavior here as something like:

```python
def clip_returns_to_benchmark(rets, benchmark_rets):
    """Drop entries from rets that precede the first entry in benchmark_rets.
    """
```

That said, is there a reason we're only clipping the left side here? Naively, I'd expect to have similar problems if rets/benchmark_rets don't align in either direction.

pyfolio/plotting.py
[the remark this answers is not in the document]
eigenfoo: Ah I see! No, that was intentional: see [this issue](https://github.com/quantopian/pyfolio/issues/534). The idea is that we want to make pyfolio as benchmark-independent as possible: if a benchmark is passed, everything works as required. If not, pyfolio simply skips the analyses that depend on benchmarks.

pyfolio/utils.py
eigenfoo: Interesting, I had just moved the code without thinking about it. The function now clips from both ends.

eigenfoo:
@twiecki @ssanderson done!

pyfolio/utils.py
twiecki: Let's remove these.

README.md
```diff
@@ -80,6 +80,42 @@ If you find a bug, feel free to [open an issue](https://github.com/quantopian/py
 You can also join our [mailing list](https://groups.google.com/forum/#!forum/pyfolio) or
 our [Gitter channel](https://gitter.im/quantopian/pyfolio).
 
+## Support
+
+Please [open an issue](https://github.com/quantopian/pyfolio/issues/new) for support.
+
+### Deprecated: Data Reading via `pandas-datareader`
+
+As of early 2018, Yahoo Finance has suffered major API breaks with no stable
```
twiecki: Rather than emphasize that the readers are deprecated, I would focus on making pyfolio independent of a benchmark being present, that's an enhancement. Then discuss how empyrical does not provide them anymore so if you need them, you have to find them.

eigenfoo: If that's the case I'd rather put that in the release notes. I'll remove the README update here.

pyfolio/utils.py
eigenfoo: done

twiecki:
Still needs to be added to the release notes.

twiecki:
Thanks @eigenfoo!

EPGM-DES:
@eigenfoo I was able to access yahoo finance historic prices using fix-yahoo-finance. However when I use price dataframe I receive the following error:

> ----pyfolio.create_returns_tear_sheet(stock_df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'pyfolio' has no attribute 'create_returns_tear_sheet'

I looked through the function in tears.py but it's unclear if there is another dependency on pandas-datareader in the function. Using fix-yahoo-finance seemed like it could be a helpful, short-term solution to the Google and Yahoo Finance API instability.

eigenfoo:
Hi @EPGM-DES, for future reference, please open up a new GitHub issue for a separate problem you're having: I directed you to this PR since it was what caused your original problem.

Without knowing more about your environment or code, it's hard for me to help you with your error. Could you open another issue with more details?

PR #543, opened 2018-06-14
MAINT: add NOT_PASSED_SENTINEL

eigenfoo:
As @ssanderson pointed out, #536 introduces a breaking change to the pyfolio API. Currently, most users are not passing a benchmark and expecting analyses to run with benchmarks. The hope is to slowly wean them off this expectation.

The solution is to introduce a separate sentinel value (in `utils.py`), which is just some sufficiently unlikely string (in this particular instance, `'__not_passed_by_user'`). If users don't pass anything, we warn them. If users pass `None` or a `pd.Series`, the analysis will run with the current behavior silently.

cc @twiecki

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -923,7 +923,10 @@ def create_round_trip_tear_sheet(returns, positions, transactions,
 
 @plotting.customize
 def create_interesting_times_tear_sheet(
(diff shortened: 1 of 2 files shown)


eigenfoo:
I'm not actually done with this PR!

twiecki:
Oops, what needs to be done?

eigenfoo:
Currently Travis fails with some indexing error, I was going to debug that and then submit for review.

Issue #556, opened 2018-08-06
IndexError: invalid index


fzhcary:
hi
I am using pyfolio 0.9.0 and pandas 0.22

Got an error below(although the program continue running and plot graphs), does anyone have ideas?

Thanks a lot!

========
IndexErrorTraceback (most recent call last)
<ipython-input-16-ca2510cddaa6> in <module>()
----> 1 pf.create_returns_tear_sheet(stock_rets, benchmark_rets=benchmark_rets)

/home/frank/Envs/quants/local/lib/python2.7/site-packages/pyfolio/plotting.pyc in call_w_context(*args, **kwargs)
     50         if set_context:
     51             with plotting_context(), axes_style():
---> 52                 return func(*args, **kwargs)
     53         else:
     54             return func(*args, **kwargs)

/home/frank/Envs/quants/local/lib/python2.7/site-packages/pyfolio/tears.pyc in create_returns_tear_sheet(returns, positions, transactions, live_start_date, cone_std, benchmark_rets, bootstrap, turnover_denom, header_rows, return_fig)
    641     if ((bootstrap is not None)
    642             and (benchmark_rets is not None)):
--> 643         ax_bootstrap = plt.subplot(gs[i, :])
    644         plotting.plot_perf_stats(returns, benchmark_rets,
    645                                  ax=ax_bootstrap)

/home/frank/Envs/quants/local/lib/python2.7/site-packages/matplotlib/gridspec.pyc in __getitem__(self, key)
    159                 raise ValueError("unrecognized subplot spec")
    160             num1, num2 = np.ravel_multi_index(
--> 161                 [_normalize(k1, nrows), _normalize(k2, ncols)], (nrows, ncols))
    162         else:  # Single key
    163             num1, num2 = _normalize(key, nrows * ncols)

/home/frank/Envs/quants/local/lib/python2.7/site-packages/matplotlib/gridspec.pyc in _normalize(key, size)
    151                 if 0 <= key < size:
    152                     return key, key
--> 153             raise IndexError("invalid index")
    154 
    155         if isinstance(key, tuple):

IndexError: invalid index

tdrobbin:
I'm also getting the same traceback with the same versions of pyfolio and pandas.

twiecki:
Thanks for the report.

twiecki:
This is a bit odd as that code shouldn't even get triggered. Can you try explicitly setting `bootstrap=False`?

twiecki:
[...]

PR #568, opened 2018-09-10
fix error when benchmark_rets is None

fehiepsi:
As titled, when `benchmark_rets=None`, `timeseries.extract_interesting_date_ranges(benchmark_rets)` will throw the error "AttributeError: 'NoneType' object has no attribute 'copy'".

This issue is also reported here: https://github.com/quantopian/pyfolio/issues/556#issuecomment-412255346

### DIFF
--- a/pyfolio/tears.py
+++ b/pyfolio/tears.py
@@ -950,8 +950,8 @@ def create_interesting_times_tear_sheet(
     if benchmark_rets is not None:
         returns = utils.clip_returns_to_benchmark(returns, benchmark_rets)
 
-    bmark_interesting = timeseries.extract_interesting_date_ranges(
-        benchmark_rets)
+        bmark_interesting = timeseries.extract_interesting_date_ranges(
+            benchmark_rets)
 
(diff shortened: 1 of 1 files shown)


pyfolio/tears.py
```diff
@@ -950,8 +950,8 @@ def create_interesting_times_tear_sheet(
     if benchmark_rets is not None:
         returns = utils.clip_returns_to_benchmark(returns, benchmark_rets)
 
-    bmark_interesting = timeseries.extract_interesting_date_ranges(
-        benchmark_rets)
+        bmark_interesting = timeseries.extract_interesting_date_ranges(
+            benchmark_rets)
```
eigenfoo: Right! I think these two lines blame to me: it's probably something I overlooked in #536.

twiecki:
Thanks @fehiepsi and @eigenfoo!

pyfolio/tears.py
shintojoseph1234: Got fixed. Thanks

Issue #570, opened 2018-09-27
Using Pyfolio to display Futures Strategies


marketneutral:
I am hoping to use `pyfolio` to create the various tearsheets on a strategy that trades futures. My first attempt via

```python
results = pd.read_pickle('../outfile.pickle')
returns, positions, transactions = pf.utils.extract_rets_pos_txn_from_zipline(results)
pf.create_simple_tear_sheet(returns, positions, transactions)
```

demonstrates that `pyfolio` doesn't know anything about contract multipliers. Concretely, I have a short position established in a strategy like

```python
    contract = data.current(context.long_bond, 'contract')
    order_target_percent(contract, -1.0)
```

i.e., a 100% of capital base short.

`pyfolio` doesn't know that there is a 1000 multiplier on this, so, for example, the "Exposure Plot" shows -0.0010 for that position.

Of course I wouldn't expect `pyfolio` to come with the data built-in, but there appears to be no way to tell `pyfolio` about these multipliers.

Is anyone using `pyfolio` to look at futures strategies? If so, how? Is anything in the works? 

As a quick fix, I would propose passing in an optional multiplier dict.

cc: @AndreasClenow

marketneutral:
On my further investigations, it appears the change would be relatively simple as the `positions` DataFrame that you pass to the various `create...tear...` functions must contain $ notional positions. As such, if we add a multiplier dict only in `utils.extract_rets_pos_txn_from_zipline(results, multipliers)` that would solve this issue I think. I am happy to make the PR on this, but would like to get some guidance if that API alteration would be acceptable, or if there are better ideas.

marketneutral:
This would solve it....

```python
from zipline.assets import Equity, Future

def apply_multipliers(positions):
    for asset in positions.columns:
        if type(asset) in [Equity, Future]:
            positions[asset] = positions[asset]*asset.price_multiplier
    return positions

results = pd.read_pickle('../outfile.pickle')
returns, positions, transactions = pf.utils.extract_rets_pos_txn_from_zipline(results)
positions = apply_multipliers(positions)
```

Any guidance on where you'd like this in the `pyfolio` code base?

IrishMarine:
Thank you both for solving this issue - I see the same issue in trying to use the create_round_trip_tear_sheet function with futures. It appears that in round_trips.py , pnl is calculated based on 'price' , which is drawn from txn.py 

Where should this be applied for correcting round_trips.py ? 

Many thanks for the help.

PR #571, opened 2018-09-29
ADD: Apply price_multiplier in Equity and Future objects to positions…

marketneutral:
… to get adjusted notionals.

Per [Issue 570](https://github.com/quantopian/pyfolio/issues/570), `pyfolio` was not applying Equity or Future multipliers. Properly defined `zipline.asset` objects `Equity` and `Future` contain meta-data, accessible with the property `.price_multiplier`, which indicates what the quoted holding size should be multiplied by to get a notional value. This PR adds this multiplication in the `extract_pos(positions, cash)` function in `pos.py`, which is called by `pf.utils.extract_rets_pos_txn_from_zipline(results)`. There is a minor change to axis labeling (removing the word "stock" in favor of "holding") and a typo fix ("Lehmann" corrected to "Lehman").

### DIFF
--- a/pyfolio/pos.py
+++ b/pyfolio/pos.py
@@ -18,6 +18,13 @@ import pandas as pd
 import numpy as np
 import warnings
 
+try:
+    from zipline.assets import Equity, Future
+    ZIPLINE = True
+except:
+    ZIPLINE = False
+    warnings.warn('Module "zipline.assets" not found; mutltipliers will not be applied to position notionals.')
+
 
 def get_percent_alloc(values):
     """
@@ -126,12 +133,18 @@ def extract_pos(positions, cash):
 
     positions = positions.copy()
     positions['values'] = positions.amount * positions.last_sale_price
+
     cash.name = 'cash'
 
     values = positions.reset_index().pivot_table(index='index',
                                                  columns='sid',
                                                  values='values')
 
+    if ZIPLINE:
+        for asset in values.columns:
+            if type(asset) in [Equity, Future]:
+                values[asset] = values[asset]*asset.price_multiplier
+    
     values = values.join(cash).fillna(0)
 
     # NOTE: Set name of DataFrame.columns to sid, to match the behavior
--- a/pyfolio/interesting_periods.py
+++ b/pyfolio/interesting_periods.py
@@ -25,7 +25,7 @@ PERIODS = OrderedDict()
 PERIODS['Dotcom'] = (pd.Timestamp('20000310'), pd.Timestamp('20000910'))
 
 # Lehmann Brothers
-PERIODS['Lehmann'] = (pd.Timestamp('20080801'), pd.Timestamp('20081001'))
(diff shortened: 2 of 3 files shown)


pyfolio/pos.py
```diff
@@ -126,12 +133,18 @@ def extract_pos(positions, cash):
...
     cash.name = 'cash'
 
     values = positions.reset_index().pivot_table(index='index',
                                                  columns='sid',
                                                  values='values')
 
+    if ZIPLINE:
+        for asset in values.columns:
+            if type(asset) in [Equity, Future]:
+                values[asset] = values[asset]*asset.price_multiplier
```
twiecki: Is this relevant also for Equities? Does `price_multiplier` always exist? Also needs a space around `*`.

marketneutral: `price_multiplier` always exists. It is a [property](https://github.com/quantopian/zipline/blob/master/zipline/assets/_assets.pyx#L75) of the base class `Asset` from which `Equity` and `Future` are derived; it is initialized to 1.0. Yes, I think this is relevant in Equities. I can see you are testing [UK Equities](https://github.com/quantopian/zipline/blob/72993a3a4e6dfb58116663f56efd1ec1bd05bed9/zipline/algorithm.py#L2389). LSE stocks are quoted in GBp (pence); i.e., 1/100 of a GBP. I believe that the GBP notional equity position would be price\*shares\*multiplier, where price is in GBp, shares are as traded, and multiplier is 0.01.

twiecki: Thanks for explaining, makes sense.

twiecki: Then just fixing the pep8 style issue and we should be good.

marketneutral: PEP8 style issue fixed.

pyfolio/pos.py
```diff
@@ -126,12 +136,18 @@ def extract_pos(positions, cash):
...
+
     cash.name = 'cash'
 
     values = positions.reset_index().pivot_table(index='index',
                                                  columns='sid',
                                                  values='values')
 
+    if ZIPLINE:
+        for asset in values.columns:
+            if type(asset) in [Equity, Future]:
```
richafrank: Since `price_multiplier` exists for all `Asset`s, what's your thought on the subtype check here? (And is it intentionally excluding subclasses of these subclasses - versus`if isinstance(asset, (Equity, Future))`?)

marketneutral: Hey @richafrank, You make a good point here. Initially I was only checking for `Future` as that was the specific case I was trying to solve for. Then I checked the `Equity` class and saw that it also had a `price_multiplier` so I simply added that. Would you recommend changing this to checking `issubclass(asset, Asset)`? To think of it, `pf.utils.extract_rets_pos_txn_from_zipline(results)` is predicated on one passing a results object from a zipline sim. As such, would it be appropriate to just assume therefore that these columns are `Asset`s and not even check?

PR #577, opened 2018-10-24
Run futurize to make code-base python 3 compatible.

twiecki:
CC @richafrank


richafrank:
@twiecki Based upon the travis build matrix, I had thought pyfolio is already py3 compatible. Does this mean that there's just not test coverage for these changed functions?

twiecki:
I was surprised too, so yes, seems like that's not tested.

twiecki:
This introduces a few problems.

PR #583, opened 2018-12-18
Dynamic setting of limit of y axis for rolling portfolio beta plot

JoeOD:
The rolling portfolio beta plot currently has a statically defined y axis limit, which ranges from -1 to 1. This sometimes results in plots like the following:

screenshot from 2018-12-17 18-33-16

Code in PR changes this to a dynamic solution, which makes plots like this: 

screenshot from 2018-12-17 18-29-22

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -884,7 +884,8 @@ def plot_rolling_beta(returns, factor_returns, legend_loc='best',
     ax.legend(['6-mo',
                '12-mo'],
               loc=legend_loc, frameon=True, framealpha=0.5)
-    ax.set_ylim((-1.0, 1.0))
+    lim = ax.get_ylim()
+    ax.set_ylim(lim)
     return ax
 
 


JoeOD:
Any pointers on what I could do to get this code to pass checks? The code I added is just a copy of code that is already in place for the drawdown periods plot, so I expected it would be fine

pyfolio/plotting.py
```diff
@@ -884,7 +884,8 @@ def plot_rolling_beta(returns, factor_returns, legend_loc='best',
     ax.legend(['6-mo',
                '12-mo'],
               loc=legend_loc, frameon=True, framealpha=0.5)
-    ax.set_ylim((-1.0, 1.0))
+    lim = ax.get_ylim()
```
twiecki: I think you can just drop this if you don't want it.

twiecki:
Thanks! The errors are not related it seems.

I think we put this in there for a reason, actually. Given that betas can be quite noisy the scale would sometimes not be in a meaningful range, so this focuses on the range we care about. It also makes that plot more easily comparable. Having said that, just fixing it statically might also not be the right approach.

twiecki:
Travis errors should be fixed in https://github.com/quantopian/pyfolio/pull/585.

JoeOD:
Ok, I see. Would it be worthwhile to try and come up with another dynamic bit of code, which is perhaps bounded in some way to keep the y-axis from getting out of control when betas get too noisy? If so, what would you be looking for conceptually?

twiecki:
@JoeOD Are you working with a high-beta portfolio?

twiecki:
One thing I could imagine is to set it to -1, 1 by default, but extend it to -2, 2 if a data point falls outside that range.

JoeOD:
Regarding the high-beta portfolio question: not really. I've just been pulling historical price data on various stocks and converting them to returns, and passing those to the tear sheet functions, just to learn how they work. I haven't seen any really noisy results so far, so I guess I haven't seen the use in truncating the range of the y-axis. I'm guessing then that idea for the plot was just to see when or if beta crosses +/- 1, without concern for how far above or below those magnitudes it goes?

gusgordon:
Closing this as it's old - looks like the diff is taken care of at this point. Feel free to reopen if this is still relevant.

PR #613, opened 2019-08-28
pep8 80 columns fixed

huaiweicheng:


huaiweicheng:
@twiecki  The test didn't pass. I have just noticed that I ran the tests below pandas=0.23.4 and all tests passed. I will change it to 0.25.1 and locate the problems.

Issue #633, opened 2020-02-26
pyfolio.timeseries.gen_drawdown_table won't work on pandas 1.0


wuzhe1234:
Pass the return series to gen_drawdown_table function,  but it won't work on pandas 1.0, and it works fine on pandas 0.25

Issue #623, opened 2019-11-03
attribute 'to_pydatetime' error


appusinghi:
How do I resolve this ?

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-35-9dcd11a12ca4> in <module>
----> 1 pf.create_full_tear_sheet(data)

C:\ProgramData\Anaconda3\lib\site-packages\pyfolio\tears.py in create_full_tear_sheet(returns, positions, transactions, market_data, benchmark_rets, slippage, live_start_date, sector_mappings, bayesian, round_trips, estimate_intraday, hide_positions, cone_std, bootstrap, unadjusted_returns, style_factor_panel, sectors, caps, shares_held, volumes, percentile, turnover_denom, set_context, factor_returns, factor_loadings, pos_in_dollars, header_rows, factor_partitions)
    209         turnover_denom=turnover_denom,
    210         header_rows=header_rows,
--> 211         set_context=set_context)
    212 
    213     create_interesting_times_tear_sheet(returns,

C:\ProgramData\Anaconda3\lib\site-packages\pyfolio\plotting.py in call_w_context(*args, **kwargs)
     50         if set_context:
     51             with plotting_context(), axes_style():
---> 52                 return func(*args, **kwargs)
     53         else:
     54             return func(*args, **kwargs)

C:\ProgramData\Anaconda3\lib\site-packages\pyfolio\tears.py in create_returns_tear_sheet(returns, positions, transactions, live_start_date, cone_std, benchmark_rets, bootstrap, turnover_denom, header_rows, return_fig)
    502                              header_rows=header_rows)
    503 
--> 504     plotting.show_worst_drawdown_periods(returns)
    505 
    506     vertical_sections = 11

C:\ProgramData\Anaconda3\lib\site-packages\pyfolio\plotting.py in show_worst_drawdown_periods(returns, top)
   1662     """
   1663 
-> 1664     drawdown_df = timeseries.gen_drawdown_table(returns, top=top)
   1665     utils.print_table(
   1666         drawdown_df.sort_values('Net drawdown in %', ascending=False),

C:\ProgramData\Anaconda3\lib\site-packages\pyfolio\timeseries.py in gen_drawdown_table(returns, top)
   1004                                                                 recovery,
   1005                                                                 freq='B'))
-> 1006         df_drawdowns.loc[i, 'Peak date'] = (peak.to_pydatetime()
   1007                                             .strftime('%Y-%m-%d'))
[...]

PR #634, opened 2020-02-26
Fixes in drawdown table when using latest version of pandas

Ildhesten:
Should adress these issues
#633 
#623

### DIFF
--- a/pyfolio/timeseries.py
+++ b/pyfolio/timeseries.py
@@ -963,7 +963,7 @@ def get_top_drawdowns(returns, top=10):
             underwater = underwater.loc[:peak]
 
         drawdowns.append((peak, valley, recovery))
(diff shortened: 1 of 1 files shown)


pyfolio/timeseries.py
```diff
@@ -963,7 +963,7 @@ def get_top_drawdowns(returns, top=10):
             underwater = underwater.loc[:peak]
 
         drawdowns.append((peak, valley, recovery))
-        if (len(returns) == 0) or (len(underwater) == 0):
+        if (len(returns) == 0) or (len(underwater) == 0) or np.min(underwater) == 0:
```
twiecki:
```suggestion
        if (len(returns) == 0) or (len(underwater) == 0) or (np.min(underwater) == 0):
```

twiecki:
Thanks @Ildhesten!

PR #674, opened 2021-06-07
[FIX]: Error when never ending drawdown

dvdpedros:
When there's just one single never ending drawdown period the code breaks when computing the drawdown plots.

### DIFF
--- a/pyfolio/plotting.py
+++ b/pyfolio/plotting.py
@@ -437,6 +437,8 @@ def plot_drawdown_periods(returns, top=10, ax=None, **kwargs):
     colors = sns.cubehelix_palette(len(df_drawdowns))[::-1]
     for i, (peak, recovery) in df_drawdowns[
             ['Peak date', 'Recovery date']].iterrows():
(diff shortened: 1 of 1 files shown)


RichardDale:
@dvdpedros I don't think Quantopian staff are involved with pyfolio/zipline/empyrical etc. any more.  They've been in radio silence mode with no comments about the future of Zipline or their involvement (some moved to Robinhood, but no idea if Zipline is part of that deal).

Thankfully it's all open source.

@stefan-jansen has done a lot of great work with Zipline (Reloaded) to make it compatible with modern versions of Python and Pandas and actually bundled it all together into a stable release.

I recommend you test against  https://github.com/stefan-jansen/pyfolio-reloaded and submit a pull request there if there's still an issue.

Cheers,
Richard.

dvdpedros:
Thanks for the detailed explanation Richard! I thought it was still maintained by people outside of the organization.
I'll try Zipline then.

Thanks,
David

PR #681, opened 2021-09-23
added index for valley in drawdown

thlautenschlaeger:
without this adaption an error got thrown since valley is an index position and not the index value


RichardDale:
Hi Thomas - Quantopian staff have been absent from this project for a while and nobody else appears to be administering it.

It has been forked here:
https://github.com/stefan-jansen/pyfolio-reloaded

and it, along with Zipline, is under active development and is compatible with newer Python/Pandas versions.

Perhaps you might want to consider submitting it there too?

Cheers,
Richard.

thlautenschlaeger:
Hi Richard,
thanks for your message I'll have a look at pyfolio-reloaded.

Best,
Thomas

